PR #10, opened 2020-07-23
pinned pymysql to 0.9.3 to ensure warnings are handled

bmildren:
##### SUMMARY
Pinning pymysql to 0.9.3, v0.10.0 changes how warnings are handled, in the interim this will ensure behaviour is as expected (specifically regarding tests), however this is in lieu of a more appropriate fix, which I'll look to follow up with.

##### ISSUE TYPE
- Bugfix Pull Request


Andersson007:
@bmildren thanks for fixing the tests!

PR #26, opened 2020-09-12
Add test to verify TLS requirements are removed

### LINKED ISSUE #20: mysql_user support of REQUIRE NONE to remove granted REQUIRE-SSL
##### SUMMARY
There is no way to remove the "REQUIRE SSL" mysql privilege ... other than rewriting the whole set of permission which is not practicable (especially when depending upon `append_privs=yes`)

Eg: After setting `REQUIRESSL`
`-  mysql_user: name=muser priv="*.*:REQUIRESSL" append_privs=yes`

... there is no way to revert such a directive. Normally this is done using REQUIRE NONE
But this is not implemented by the Ansible role.

`-  mysql_user: name=muser priv="*.*:REQUIRENONE" append_privs=yes`

Expected result: The REQUIRE SSL is droped for _muser_

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
mysql_user

##### ADDITIONAL INFORMATION
https://github.com/ansible/ansible/issues/15440
https://github.com/ansible/ansible/issues/29811

Jorge-Rodriguez:
##### SUMMARY
Add integration tests to verify that an empty `tls_requires` object removes TLS restrictions on client connection.

Fixes #20 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
mysql_user


Jorge-Rodriguez:
@Andersson007 There's a weird side effect causing the integration  error. Can we trigger the run again? I can't figure out why the user update in check mode for the resource limits detects a change on the TLS requirements.

```
TASK [test_mysql_user : Try to set the same limits again in check mode] ********
changed: [testhost] => {"changed": true, "msg": "TLS requires updated", "user": "db_user1"}

TASK [test_mysql_user : assert] ************************************************
fatal: [testhost]: FAILED! => {
    "assertion": "result is not changed",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

Andersson007:
@bmildren my local tests fail with:
```
TASK [setup_mysql | config | setup replication topology] ***********************
task path: /root/ansible_collections/community/mysql/tests/output/.tmp/integration/test_mysql_user-ok_yhfad-ÅÑŚÌβŁÈ/tests/integration/targets/setu
p_mysql/tasks/config.yml:12
fatal: [testhost]: FAILED!
```
this part was introduced with
```
[root@spblnx124 mysql]# git blame -L 14,14 tests/integration/targets/setup_mysql/tasks/config.yml 
ac79b437 (Ben Mildren 2020-07-14 14:13:16 +0100 14)     cmd: "dbdeployer deploy multiple {{ install_version }} --base-port {{ mysql_base_port }}
```

Andersson007:
i use `ansible-test integration test_mysql_user --docker ubuntu1804 --docker-no-pull -vv > ~/test.log`
maybe there's another way to run them without catching the error above?

Andersson007:
@Jorge-Rodriguez unfortunately there's no way to rerun only failed stuff, so reruned all

bmildren:
@Andersson007 could you try without specifying ubuntu1804 as the test container?  e.g. `ansible-test integration test_mysql_user --docker -vv > ~/test.log`.

This would default to using [default-test-container](https://github.com/ansible/default-test-container/blob/master/Dockerfile), this is still based on ubuntu1804, but has several additional packages installed.

Andersson007:
@bmildren it works, thanks
Can we run the tests on different distributions or they now cover all supported mysql versions? Also can't some things depend on a certain distribution?
If the default container is enough, we should definitely add a note how to run integration tests locally (in README.md?) for newcomers.

bmildren:
> If the default container is enough, we should definitely add a note how to run integration tests locally (in README.md?) for newcomers.

Yep definitely makes sense I'll add that.

> Can we run the tests on different distributions or they now cover all supported mysql versions? Also can't some things depend on a certain distribution?

Currently it's only MySQL Server, but we can cover all the different distributions (Percona, MariaDB, MySQL), plus different topogies such as galera if we choose (dbdeployer covers all options).  I can enable MariaDB (with the minor downside it'll extend the duration) and yes we can target certain tests based on distribution 👍

Andersson007:
@bmildren cool, thanks!

Andersson007:
https://github.com/ansible-collections/community.mysql/issues/27 as a reminder

Jorge-Rodriguez:
OK, there's a bug in the way we parse the Create User response to check the TLS parameters. Working on a fix.

Andersson007:
@Jorge-Rodriguez thanks, looks like a change:)
Several questions:
1. Can the changes affect current user playbooks behavior anyhow? Any possible and impossible cases.
2. You just moved CI tests to a separate file, added for your case, not changing current ones?
3. Do we have total coverage of TLS related stuff in the existing tests?

I'm not a user of this (very important) module, so i'm asking to be sure we don't introduce any breaking / unexpected changes in the module's behavior.

Jorge-Rodriguez:
@Andersson007 the changes in this PR are bugfixes for the `tls_requires` functionality so it won't affect playbooks that do not use that parameter and it will correct unexpected behaviours in those that do. Since we don't have bug reports regarding that, I assume there aren't many playbooks using the functionality yet.

The CI tests have been moved and updated. We're now also checking that we can remove the TLS requirements by setting `tls_requires` to `None`, an example is provided in the documentation.

I would say we do have total coverage. We're checking all CRUD use cases plus the resource limits tests verify t as a side effect hat there aren't any interactions with other requirements.

Andersson007:
@Jorge-Rodriguez thanks for fixing this!

Andersson007 (approved):

Andersson007:
@Jorge-Rodriguez we forgot a changelog fragment! Could you please add this ASAP in a separate PR? We'll release the collection next week.

PR #32, opened 2020-09-24
CI tests: add a note not to use tests as an example of writing roles

Andersson007:
##### SUMMARY
CI tests: add a note not to use tests as an example of writing roles

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
CI tests


Andersson007:
because they are included in release tarballs. We discuss this yesterday at community meeting in IRC

PR #36, opened 2020-09-27
Do not run same sanity tests twice

felixfontein:
##### SUMMARY
The sanity tests for every Ansible version are run twice, because the Python version used to execute ansible-test does not matter (since `--docker` is used).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
CI


Andersson007 (approved):
LGTM, thanks @felixfontein!

PR #41, opened 2020-09-30
Remove parsing of REQUIRE SSL in privileges

baszoetekouw:
This fixes a bug in which a `mysql_user` entry with a `tls_requires` option would always be marked as changed.   

By adding the `REQUIRESSL` option to the current privileges, it gets passed to to the `curr_priv` variable in `user_mod()`, causing the privileges intersection between `curr_priv` and `new_priv` on line 663 to always contain `REQUIRESSL` (because `new_priv` will never never contain `REQUIRESSL` it if we use `tls_requires`).  Therefore the module would always be marked as "changed":  it would always try to remove the REQUIRESSL via `privileges_revoke()` at line 669 and then add it back via the explicit TLS route at lines 673-690).

I'm aware this might not be the optimal solution.  Maybe it would be better to fix the ordering in `user_mod()` by first checking if `tls_requires` implies REQUIRE SSL, if so, add it to `new_priv` (if not already present), and then proceed with modifying the privileges and TLS settings.  However, that might defeat the purpose of having a separate `tls_requires` option in the first place.

### DIFF
--- a/plugins/modules/mysql_user.py
+++ b/plugins/modules/mysql_user.py
@@ -746,8 +746,6 @@ def privileges_get(cursor, user, host):
         privileges = [pick(x.strip()) for x in privileges]
         if "WITH GRANT OPTION" in res.group(7):
             privileges.append('GRANT')
-        if 'REQUIRE SSL' in res.group(7):
-            privileges.append('REQUIRESSL')
         db = res.group(2)
         output.setdefault(db, []).extend(privileges)
     return output


Jorge-Rodriguez:
@baszoetekouw if I understand correctly, this change would transfer the issue from a task that has both REQURESSL in `privs` and the `require_tls` option, which while technically possible should not happen (there's a note about that in the documentation) to a task that only has the REQUIRESSL `priv` which is a valid and backwards compatible task. If we stop parsing the SSL requirement, a task with only the REQUIRESSL priv will never detect that the requirement is set and so, it will always show as changed.

In a perfect world, I would deprecate REQUIRESSL in `priv` and drop support for it altogether. but as long as we need to keep backwards compatibility we're stuck with this.

Personally I would not merge this as it fixes an issue that's ultimately caused by bad task writing.

Ping @Andersson007

baszoetekouw:
No, sorry for not being clearer.   The issue is that a task with _only_ a "modern" `tls_requires` option (and no REQUIRESSL in `priv`) always returns `changed: true`.

In particular,  the following task keep returning "changed: true" (for a MariaDB 10.3):
```
- name: Add database user
  mysql_user:
    login_host: '127.0.0.1'
    login_user: 'root'
    login_password: v3ry_'s3cr1t'
    name: 'test_user'
    host: '127.0.0.1'
    password: 's3cr1t'
    priv: 'some_db.*:ALL'
    tls_requires: { 'SSL' }
```

The issue seems to be that MariaDB (not sure if MySQL behaves in the same way) always returns `REQUIRE SSL` as part of the grants, even if we set the `tls_requires` as in the fragment above:
```
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 108
Server version: 10.3.23-MariaDB-0+deb10u1 Debian 10

Copyright (c) 2000, 2018, Oracle, MariaDB Corporation Ab and others.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

MariaDB [(none)]> SHOW GRANTS FOR `test_user`@`127.0.0.1`;
+------------------------------------------------------------------------------------------------------------------------------+
| Grants for test_user@127.0.0.1                                                                                               |
+------------------------------------------------------------------------------------------------------------------------------+
| GRANT USAGE ON *.* TO `test_user`@`127.0.0.1` IDENTIFIED BY PASSWORD '*EB3AE1389A7D5489888C172CBA962B8710190B9C' REQUIRE SSL |
| GRANT ALL PRIVILEGES ON `some_db`.* TO `test_user`@`127.0.0.1`                                                               |
+------------------------------------------------------------------------------------------------------------------------------+
2 rows in set (0.002 sec)
```

This causes the following chain of events:
  - the module adds `REQUIRESSL` to `curr_priv` in https://github.com/ansible-collections/community.mysql/blob/f6d552d95e50e1a4fee5e0c008a3386639d28d3c/plugins/modules/mysql_user.py#L749
 - `REQUIRESSL` is not in `new_priv` (because we _don't_ put it the `priv` option in the task)
 - the module detects a difference between table privileges at https://github.com/ansible-collections/community.mysql/blob/f6d552d95e50e1a4fee5e0c008a3386639d28d3c/plugins/modules/mysql_user.py#L661 
 - `REQUIRESSL` is _removed_ at https://github.com/ansible-collections/community.mysql/blob/f6d552d95e50e1a4fee5e0c008a3386639d28d3c/plugins/modules/mysql_user.py#L669
 - `current_requires` will now be `None`  at https://github.com/ansible-collections/community.mysql/blob/f6d552d95e50e1a4fee5e0c008a3386639d28d3c/plugins/modules/mysql_user.py#L674 (since `REQUIRESSL` has just been removed)
 - the module with re-add the TLS requirement at https://github.com/ansible-collections/community.mysql/blob/f6d552d95e50e1a4fee5e0c008a3386639d28d3c/plugins/modules/mysql_user.py#L686

So the problem really is that even if we only use `tls_requires`, the module still mixes REQUIRESSL into the privileges.

Jorge-Rodriguez:
Thanks for the detailed information. I'm going to have a look at the codebase and make sure this is handled properly, then I'll double check this PR

Jorge-Rodriguez:
Off of the top of my head, I think it would be better to make that parsing conditional on the `tls_requires` option not being `None`, that way the old behaviour would remain unchanged

plugins/modules/mysql_user.py
```diff
@@ -746,8 +746,6 @@ def pick(x):
         privileges = [pick(x.strip()) for x in privileges]
         if "WITH GRANT OPTION" in res.group(7):
             privileges.append('GRANT')
```
Jorge-Rodriguez, line 748:
```suggestion
            privileges.append('GRANT')
        if 'REQUIRESSL' in res.group(7) and not tls_requires:
            privileges.append('REQUIRESSL')
```

Jorge-Rodriguez: Mind that `tls_requires` is not defined in this context

Jorge-Rodriguez (changes requested):

plugins/modules/mysql_user.py
baszoetekouw: Yeah, I think that makes sense.  I'll try to come up with on updated patch.

Jorge-Rodriguez: When you update this, can you also add tests to verify the behaviour, both in tasks with `priv` and with `tls_requires`?

Andersson007:
@baszoetekouw thanks for the PR!

Could you also please add:
1) a changelog fragment https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to
2) integration tests, see content of tests/integration/targets/test_mysql_user/

Andersson007:
any updates on this?

baszoetekouw:
Sorry, haven't gotten around to this yes; I'll try to make some time in the next week.

Andersson007:
@baszoetekouw thanks!

Andersson007:
@baszoetekouw hi, maybe if there are time issues or something, wouldn't you mind if anybody else finish the PR?

Jorge-Rodriguez:
@baszoetekouw @Andersson007 I think we may want to put this PR on hold, pending the discussion in #101

Andersson007:
@Jorge-Rodriguez sounds good

Jorge-Rodriguez:
@baszoetekouw If you didn't follow #101 it appears we're getting rid of the REQUIRESSL privilege altogether. Do you want to tackle that change yourself along with this?

Andersson007:
if no response, IMO, we should close the PR then, say, on Friday (we could always open it again if needed).

baszoetekouw:
Sorry about not responsing sooner.  I'm pretty swamped right now, so feel free to close this PR for now.  I can retest my setup (and possible come up with a new PR) once #101 has landed.

Andersson007:
@baszoetekouw ok, thanks for the feedback!

PR #63, opened 2020-11-21
Add missing options to CHANGE MASTER TO

### LINKED ISSUE #59: mysql_replication: make MASTER_SSL_VERIFY_SERVER_CERT available
##### SUMMARY
The _mysql_replication_ module should support setting **MASTER_SSL_VERIFY_SERVER_CERT**.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
mysql_replication (changemaster)

##### ADDITIONAL INFORMATION
In an environment with a central CA as a trust anchor it certainly makes sense to verify the servers certificate. Otherwise a single compromised certificate+key would enable a MitM attack against the replication connection. Security minded people like to configure the replication connection in the most secure way possible.

Jorge-Rodriguez:
##### SUMMARY
Added all missing options for CHANGE MASTER TO, according to https://dev.mysql.com/doc/refman/8.0/en/change-master-to.html

Fixes #59 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
mysql_replication


Jorge-Rodriguez:
@Andersson007 @bmalynovytch @bmildren I'm not quite sure how to write tests for this, other than running the module with the new options and check that the sent query matches expectation and the server doesn't spit an error. Do you have any suggestions?

Jorge-Rodriguez:
@Andersson007 the sanity checks say I can't define choices for an integer parameter. What's the proper way to define an `int` within a specific range of values?

changelogs/fragments/59-mysql_replication_verify_server_cert.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- mysql_replication - add missing CHANGE MASTER TO options (https://github.com/ansible-collections/community.mysql/issues/59).
```
Andersson007:
```suggestion
- "mysql_replication - add the new options: ``master_bind``,... (https://github.com/ansible-collections/community.mysql/issues/59)."
```
alternatively, we could use separate entries for each parameter

plugins/modules/mysql_replication.py
```diff
@@ -42,6 +43,11 @@
     - resetslave
     - resetslaveall
     default: getslave
+  master_bind:
+    description:
+    - same as mysql_variable.
```
Andersson007:
```suggestion
    - same as mysql_variable.
```
```suggestion
    - Same as mysql variable.
```

Andersson007:
> @Andersson007 the sanity checks say I can't define choices for an integer parameter. What's the proper way to define an `int` within a specific range of values?

I've never seen int ranges, there aren't a lot to list the whole sequence, i.e. [1, 2, ...] , or maybe we could just document this and check in the code if a passed value is in the range.

Andersson007:
> @Andersson007 @bmalynovytch @bmildren I'm not quite sure how to write tests for this, other than running the module with the new options and check that the sent query matches expectation and the server doesn't spit an error. Do you have any suggestions?

Can we run the 3d instance? As i can see there are `mysql_primary_port` and `mysql_replica1_port` there now.
Maybe to run one more as a replica with `mysql_replica2_port` and switch it to `mysql_replica1_port` as a master.
I'm not sure if it works with MySQL but in Postgres we can use a "cascade" of replicas and switch between them so that one will become a master to another being a replica of another one without rebuilding.
If it's complicated, your suggestion, imo, also works.

but @bmildren and @bmalynovytch , as experts, know better:)

bmalynovytch:
Absolutely, you may have 1 master with 2 slaves and switch one of the slaves to master and attach the second one as new slave of the new master.

plugins/modules/mysql_replication.py
Jorge-Rodriguez: Argh, IDE autocomplete. Nicely spotted. 👍

Jorge-Rodriguez:
@bmalynovytch since all of the options to `CHANGE MASTER TO` are optional and we aren't testing the `MASTER_HOST` or `MASTER_PORT` options (those are already tested) couldn't we test everything against the running master?

The other, more complicated question is what kind of effects to look for to verify that the command succeeded. But in that case I would assume that if the query sent to the server matches the expectation and the server doesn't err then the server has done as requested.
It could be the case that there server remains silent from the module's PoV but didn't honour the request, but my knowledge of mysql replication is not deep enough to foresee these cases.

Jorge-Rodriguez:
> > @Andersson007 the sanity checks say I can't define choices for an integer parameter. What's the proper way to define an `int` within a specific range of values?
> 
> I've never seen int ranges, there aren't a lot to list the whole sequence, i.e. [1, 2, ...] , or maybe we could just document this and check in the code if a passed value is in the range.

Maybe we can define it as a list of string representations of integers. At the end of the day however the value is written in YAML didn't make that much of a difference, and we end up turning it into a string for the cursor query anyway.

Andersson007:
> > > @Andersson007 the sanity checks say I can't define choices for an integer parameter. What's the proper way to define an `int` within a specific range of values?
> > 
> > 
> > I've never seen int ranges, there aren't a lot to list the whole sequence, i.e. [1, 2, ...] , or maybe we could just document this and check in the code if a passed value is in the range.
> 
> Maybe we can define it as a list of string representations of integers. At the end of the day however the value is written in YAML didn't make that much of a difference, and we end up turning it into a string for the cursor query anyway.

For me personally the best approach is documenting the acceptable range, removing choices, and checking on the passed value in the code with raising a helpful message if the value is out of the borders.

Jorge-Rodriguez:
I'm going to abandon this PR draft, as it relates to replication and we have modifications regarding that component and the language used.
That being said, I'd like to keep this PR still open for reference until I get the chance to tackle the issue again.

Andersson007:
@Jorge-Rodriguez sure, it's marked as DRAFT, so, people will be warned. Maybe also would be good to put something like "TEMPORARY FROZEN" to the description.

Andersson007:
@Jorge-Rodriguez any updates on the PR?

Jorge-Rodriguez:
I haven't gotten the chance to work on this for quite some time. I'll pick it up, seems there's only the "enum" issue to solve and to write appropriate tests.

Thanks for the ping.

Jorge-Rodriguez:
@Andersson007 I have just brought the code of this PR back to date. Still pending changes to the "enum" issue and testing, so I'm not moving it from draft status yet.

Jorge-Rodriguez:
I am unable to run integration tests locally, When running on docker the first `get_url` task fails with HTTP 403. The URL is otherwise OK and it works, so it would seem an issue with docker?

Jorge-Rodriguez:
I found out that all our tests were run with the `install_mariadb` flag set to true.

Jorge-Rodriguez:
Tests are broken because downloading the MySQL tarball consistently fails with HTTP 403. The reason is unknown as of this writing.

Jorge-Rodriguez:
It seems that the old `dev.mysql.com` URL is not working anymore and the resources have been moved to `cdn.mysql.com`. For some reason, the request from the `get_url` module does not get redirected.

Andersson007:
@Jorge-Rodriguez great work, thank you!
And thanks for investigating the reason of 403.

1. Would you like to raise a separate PR for the URL change you did here?
2. Do you think we need those `master/slave` aliases? I think we left them for older options not to break stuff in users WFs. As these are new options, not sure if we should add the aliases. Ideas?

Jorge-Rodriguez:
Yeah, it might be a good idea not to introduce futile backwards compatibility on new features, I'll remove the aliases.

Andersson007:
@Jorge-Rodriguez thanks! please let me know when it's ready for review.
There are a lot of parameters. Are they all supported by all MySQL/MariaDB supported versions?
If not, i'm not sure if we should document minimal supporting versions for each new argument or not? Maybe it's an extra thing. Ideas?

Jorge-Rodriguez:
I've just rebased the whole thing to make the commit history clearer. I'm still keeping the CI fixup so the tests run correctly, I'll remove them when #491 is merged.

@Andersson007 no need for review just yet. I still have to write tests for the new parameters, I'm going to keep the PR as a draft until those are ready.

Jorge-Rodriguez:
@Andersson007 What's the proper way to document an option that is only available since a specific version of MySQL?

Andersson007:
> @Andersson007 What's the proper way to document an option that is only available since a specific version of MySQL?

@Jorge-Rodriguez imo just to add another row to each option saying something like `- Available (or supported maybe) since MySQL <version > / MariaDB <version>` would be enough.

@laurent-indermuehle @betanummeric @bmalynovytch other thoughts ^ ?

laurent-indermuehle:
1) If used on older version, will it be ignored or crash?
2) if used on another engine (mysql vs mariadb), will it be ignored or crash?

I hope that 'Plugins CI' will tell us once it runs ;)

Andersson007:
> 1. If used on older version, will it be ignored or crash?
> 2. if used on another engine (mysql vs mariadb), will it be ignored or crash?
> 
> I hope that 'Plugins CI' will tell us once it runs ;)

I believe just crashing is OK as soon as the versions are documented. We, of course, can check versions in our code, etc. but shouldn't this be user's responsibility? If a user runs unsupported SQL, RDBMS just shows errors (invalid syntax or something), doesn't it?

laurent-indermuehle:
> I believe just crashing is OK as soon as the versions are documented. We, of course, can check versions in our code, etc. but shouldn't this be user's responsibility? If a user runs unsupported SQL, RDBMS just shows errors (invalid syntax or something), doesn't it?
Right, ansible-doc will show the user that we documented the minimum version and his logs will tell him what went wrong.

I searched the repo and found many occurences:

- Supported from MariaDB 10.0.1.
- Multi-source replication is supported from MySQL 5.7.
- Roles are supported since MySQL 8.0.0 and MariaDB 10.0.5.
- Supported by B(MariaDB).
- Is not supported by MariaDB and is silently ignored when working with MariaDB.
- Supported by MySQL 8.0 or later.

English is not my first language, so I don't know what the most correct form is. But I know I don't like seen a dot at the end of the version.

Jorge-Rodriguez:
I'm planning to spot the problematic options as I progress with test writing and have the module fail gracefully with an appropriate message.

Jorge-Rodriguez:
In a different order of things, it seems that since we started making modificaciones to the replication module, MySQL has settled on replacing the `master` term with `source` instead of our chosen `primary` so while I'm at this I'm going to introduce that rename too

Jorge-Rodriguez:
> > 1. If used on older version, will it be ignored or crash?
> > 2. if used on another engine (mysql vs mariadb), will it be ignored or crash?
> > 
> > I hope that 'Plugins CI' will tell us once it runs ;)
> 
> I believe just crashing is OK as soon as the versions are documented. We, of course, can check versions in our code, etc. but shouldn't this be user's responsibility? If a user runs unsupported SQL, RDBMS just shows errors (invalid syntax or something), doesn't it?

The problem with delegating the error to the connector is that it can become very unclear. As a user you'd get a message such as 

```
[...] PLEASE CHECK YOUR SYNTAX NEAR [...]
```
But the user didn't construct that query, the module did, so it would require intimate knowledge of the module's inner workings to determine what is causing the problem.

IMO it'd be better to have the module do the version checking and fail with a proper message like
```
You're using DB version A and argument 'you_jumped_the_gun' isn't supported until version B.
```

Jorge-Rodriguez:
> > I believe just crashing is OK as soon as the versions are documented. We, of course, can check versions in our code, etc. but shouldn't this be user's responsibility? If a user runs unsupported SQL, RDBMS just shows errors (invalid syntax or something), doesn't it?
> > Right, ansible-doc will show the user that we documented the minimum version and his logs will tell him what went wrong.
> 
> I searched the repo and found many occurences:
> 
> * Supported from MariaDB 10.0.1.
> * Multi-source replication is supported from MySQL 5.7.
> * Roles are supported since MySQL 8.0.0 and MariaDB 10.0.5.
> * Supported by B(MariaDB).
> * Is not supported by MariaDB and is silently ignored when working with MariaDB.
> * Supported by MySQL 8.0 or later.
> 
> English is not my first language, so I don't know what the most correct form is. But I know I don't like seen a dot at the end of the version.

It's a linguistics problem because we're mixing the syntaxes of versioning and the English language, the latter being infamous for being anything but formal. We could of course do some language gymnastics and rewrite it as something like
> Supported from version 10.0.5 of MariaDB.

although it's a bit clunky. The last option is probably best since it keeps the same semantics while removing the version number from the end of the sentence
> Supported by MySQL version 8.0 and later.

We even get the chance to use Oxford commas
> Supported by MySQL version 8.0.0 and later, and MariaDB version 10.0.5 and later.

Or, we leverage markdown notation
```
Supported since MySQL version `8.0`.
```

Andersson007:
@Jorge-Rodriguez 

>In a different order of things, it seems that since we started making modificaciones to the replication module, MySQL has settled on replacing the master term with source instead of our chosen primary so while I'm at this I'm going to introduce that rename too

I hope MariaDB will not introduce other aliases, there'll be too many imo in that case:))

>IMO it'd be better to have the module do the version checking and fail with a proper message like

if you're happy to write all the checks, WFM:)

>Supported by MySQL 8.0 or later.

I like this one ^. Plus I would add the markdown to highlight the versions a bit, so something like:

>Supported by MySQL C(8.0) or later

Though I would suggest to conduct the rewording in a separate PR:)
I'm a strong supporter of the principle `Split all that can be split into separate PRs`

Jorge-Rodriguez:
MariaDB od still stuck in the political correctness flame war of 2020, so I don't expect much progress on the rewording front any time soon

laurent-indermuehle:
Since MariaDB 10.5 you can write `show replica status\G` instead of `show slave status\G`. So, we may need a compatibility table to keep up :/

Jorge-Rodriguez:
> Since MariaDB 10.5 you can write `show replica status\G` instead of `show slave status\G`. So, we may need a compatibility table to keep up :/

Well, that would be in line with MySQL terms.

Jorge-Rodriguez:
This is still in progress, I want to be thorough with the test suite so it's taking some time.

It seems that MariaDB is going with `primary` instead of `source`, but they still haven't made changes to the change master command or its options.

I will introduce support for MySQL's `change replica source` and its options but I'll leave the new aliases for a separate PR.

Andersson007:
> This is still in progress, I want to be thorough with the test suite so it's taking some time.
> 
> It seems that MariaDB is going with `primary` instead of `source`, but they still haven't made changes to the change master command or its options.
> 
> I will introduce support for MySQL's `change replica source` and its options but I'll leave the new aliases for a separate PR.

@Jorge-Rodriguez are we talking about reflecting the differences in naming in aliases? If yes, i wonder if we really need them. Maybe we should just use one set of names? Maybe we could just mention in `notes:` that, say, options containing `primary` can mean `source` in some versions of the RDBMSes?

Jorge-Rodriguez:
> > This is still in progress, I want to be thorough with the test suite so it's taking some time.
> > It seems that MariaDB is going with `primary` instead of `source`, but they still haven't made changes to the change master command or its options.
> > I will introduce support for MySQL's `change replica source` and its options but I'll leave the new aliases for a separate PR.
> 
> @Jorge-Rodriguez are we talking about reflecting the differences in naming in aliases? If yes, i wonder if we really need them. Maybe we should just use one set of names? Maybe we could just mention in `notes:` that, say, options containing `primary` can mean `source` in some versions of the RDBMSes?

While not strictly necessary, I believe the option aliases would work better in the future for users who are used to the newer terminology. I also think an alias would probably work better than an explanation in notes that might not be read.
I'm any case none of this will concern this particular PR.

Andersson007:
> > > This is still in progress, I want to be thorough with the test suite so it's taking some time.
> > > It seems that MariaDB is going with `primary` instead of `source`, but they still haven't made changes to the change master command or its options.
> > > I will introduce support for MySQL's `change replica source` and its options but I'll leave the new aliases for a separate PR.
> > 
> > 
> > @Jorge-Rodriguez are we talking about reflecting the differences in naming in aliases? If yes, i wonder if we really need them. Maybe we should just use one set of names? Maybe we could just mention in `notes:` that, say, options containing `primary` can mean `source` in some versions of the RDBMSes?
> 
> While not strictly necessary, I believe the option aliases would work better in the future for users who are used to the newer terminology. I also think an alias would probably work better than an explanation in notes that might not be read. I'm any case none of this will concern this particular PR.

I'm not strictly against of the aliases if it helps though i think we shouldn't overuse them:) OK, let's discuss this later in a dedicated PR/Issue/Discussion

simonpahl:
I would like to use the MASTER_BIND with the module, would it be acceptable to extract parts of this PR to support this parameter?

laurent-indermuehle:
Hi @simonpahl, I think it's a question for @Jorge-Rodriguez.
But if he doesn't respond, I think it would be ok to create a new PR based on his work.
It's an old PR and probably needs a lot of rebasing anyway.

PR #71, opened 2020-12-09
mysql_replication: fix crashes caused by deprecated terminology

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.mysql/issues/70

mysql_replication: fix crashes caused by deprecated terminology

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
`plugins/modules/mysql_replication.py`

PR #72, opened 2020-12-12
mysql_user: Fixed change detection with append_privs (#69)

### LINKED ISSUE #69: append_privs attempts to change privileges when it doesn't need to
##### SUMMARY
append_privs usage will result in detecting a change to permissions even when the set of current privileges is a superset of new privileges.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
mysql_user

##### ANSIBLE VERSION

```
ansible 2.10.3
  config file = None
  configured module search path = ['/home/$USER/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/$USER/.local/share/virtualenvs/ansible-testing-3MbEpyMU/lib/python3.8/site-packages/ansible
  executable location = /home/$USER/.local/share/virtualenvs/ansible-testing-3MbEpyMU/bin/ansible
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```

##### CONFIGURATION
```
no output (isolated test environment)
```

##### OS / ENVIRONMENT
This doesn't seem to be specific to the environment in any way.

##### STEPS TO REPRODUCE
1. `docker run -p 3306:3306 --name mysql -e MYSQL_ROOT_PASSWORD=password -d mysql:5.7`
2. `create database test; use test; create table users (id int primary key);`
3. `ansible-playbook mysql.yml`
4. Run the command above for a second time and confirm that the plugin doesn't detect any changes
5. Remove the INSERT privileges from the play and run the playbook again.  The plugin detects that a change was made.  Continue to run this any number of times and a change will always be detected.

```yaml
---

- name: Test MySQL
  gather_facts: false
  hosts: localhost
  tasks:

    - name: Create database user with privileges
      mysql_user:
        login_user: root
        login_password: password
        login_host: 127.0.0.1
        name: bob
        password: '12345'
        host: '%'
        priv: 'test.users:SELECT,INSERT'
        append_privs: true
        state: present
```

##### EXPECTED RESULTS
The expected behavior is that the current privileges (SELECT,INSERT) should not require a change if the new permissions are SELECT.

##### ACTUAL RESULTS
The play will always report a change because the current permissions (MySQL grant tables) and new permissions (defined in Ansible play) will never match.

I am not ruling out that this could be the desired behavior.  What isn't clear to me is whether the purpose of `append_privs` is to avoid the REVOKE statement (which can result in outage), or whether it is to indicate that the only concern is that the current privileges are a superset of the new privileges.  The current behavior is not truly reporting whether change is happening to the user or not.  Additionally, it's running grants that could be skipped.

A simple implementation that would fix this particular issue (requires more testing) is:

```python
                # If appending privileges, only the permissions that exist in the new privileges, but not the current
                # privileges matter (difference vs. symmetric difference).
                if append_privs:
                    priv_diff = set(new_priv[db_table]) - set(curr_priv[db_table])
                else:
                    priv_diff = set(new_priv[db_table]) ^ set(curr_priv[db_table])
```

Some examples of expected behavior using this code:

```
current: SELECT
new: SELECT,INSERT
result: change

current: SELECT,INSERT
new: DELETE
result: change

current: SELECT,INSERT
new: SELECT
result: unchanged
```

This change would definitely require more testing for edge cases that I've missed.  I'm happy to run with this because there might be some additional changes that I'd like to contribute in the near future.

If the current behavior is the desired behavior, it might be helpful to change the documentation on append_privs to include:

* intent of this parameter (to avoid the potential for outage vs. ensuring current privileges are at least as permissive as new permissions)
* note that a change will always be detected if the permissions do not match exactly

Thank you for the work on this plugin!

steveteahan:
Prior to this change, mysql_user with append_privs would attempt to make
a change even if the current privileges were a superset of the new
privileges (shouldn't require any action).

##### SUMMARY

Fixes #69 - mysql_user append_privs change detection where current privileges are a superset of new privileges

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
mysql_user

##### ADDITIONAL INFORMATION

The issue is detailed in #69.

### DIFF
--- a/tests/integration/targets/test_mysql_user/tasks/test_priv_append_no_change.yml
+++ b/tests/integration/targets/test_mysql_user/tasks/test_priv_append_no_change.yml
@@ -44,6 +44,7 @@
         priv: 'data1.*:SELECT/data2.*:SELECT'
         append_privs: yes
         state: present
+      check_mode: '{{ enable_check_mode }}'
       register: result
 
     - name: Assert that there wasn't a change and that the permissions are still the same
--- a/tests/integration/targets/test_mysql_user/tasks/test_priv_append_no_change.yml
+++ b/tests/integration/targets/test_mysql_user/tasks/test_priv_append_no_change.yml
@@ -1,0 +1,79 @@
+# Test code to ensure that appending privileges will not result in unnecessary changes when the current privileges
+# are a superset of the new privileges that have been defined.
+- vars:
+    mysql_parameters: &mysql_params
+      login_user: '{{ mysql_user }}'
+      login_password: '{{ mysql_password }}'
+      login_host: 127.0.0.1
+      login_port: '{{ mysql_primary_port }}'
+
+  block:
+
+    - name: Create test databases
+      mysql_db:
+        <<: *mysql_params
+        name: '{{ item }}'
+        state: present
+      loop:
+      - data1
+      - data2
+
+    - name: Create user with privileges
+      mysql_user:
+        <<: *mysql_params
+        name: '{{ user_name_4 }}'
+        password: '{{ user_password_4 }}'
+        priv: 'data1.*:SELECT,INSERT/data2.*:SELECT,DELETE'
+        state: present
+
+    - name: Run command to show privileges for user (expect privileges in stdout)
+      command: "{{ mysql_command }} -e \"SHOW GRANTS FOR '{{ user_name_4 }}'@'localhost'\""
+      register: result
+
+    - name: Assert user given privileges
+      assert:
+        that:
+          - "'GRANT SELECT, INSERT ON `data1`.*' in result.stdout"
+          - "'GRANT SELECT, DELETE ON `data2`.*' in result.stdout"
+
+    - name: Append privileges that are a subset of the current privileges, which should be a no-op
+      mysql_user:
+        <<: *mysql_params
+        name: '{{ user_name_4 }}'
+        password: '{{ user_password_4 }}'
+        priv: 'data1.*:SELECT/data2.*:SELECT'
+        append_privs: yes
+        state: present
+      register: result
+
+    - name: Assert that there wasn't a change and that the permissions are still the same
+      assert:
+        that:
+          - "result.changed == false"
+
+    - name: Run command to show privileges for user once more (expect privileges in stdout)
+      command: "{{ mysql_command }} -e \"SHOW GRANTS FOR '{{ user_name_4 }}'@'localhost'\""
+      register: result
+
+    - name: Assert user given privileges once more
+      assert:
+        that:
+          - "'GRANT SELECT, INSERT ON `data1`.*' in result.stdout"
+          - "'GRANT SELECT, DELETE ON `data2`.*' in result.stdout"
+
+    ##########
+    # Clean up
+    - name: Drop test databases
+      mysql_db:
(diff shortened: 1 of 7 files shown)


Andersson007 (commented):
@steveteahan thanks for the PR!
Also, please,
1) create a changelog fragment https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to
2) fix sanity errors to make CI green
3) add a test case covering `check_mode: yes`

steveteahan:
I can take care of 1 and 3 no problem.  I was actually going to touch base on the CI errors because they seem to be related to mysql_replication argument_specs.  It looks like this is actually broken in the main branch as well, so it seems to be unrelated to this change.  I'll get the fragments and check mode covered first and then see if there's anything else I can figure out on the sanity errors.

steveteahan:
It looks like the same issue was fixed in: https://github.com/ansible-collections/community.mysql/pull/71/commits/81b3670e11bfcc453ded94e8460e00ebb056f8ae

I'll go ahead and add it to this PR.

Andersson007:
@steveteahan thanks for this!
Could you please also add a test case expanding privileges for one of them, e.g.:
priv: 'data1.*:DELETE/data2.*:SELECT', then check

These changes look logical to me but i myself am not using the module.
While you're adding the case I'll email our WG leader to look at the PR

tests/integration/targets/test_mysql_user/tasks/test_priv_append_no_change.yml
```diff
@@ -0,0 +1,80 @@
...
+        that:
+          - "'GRANT SELECT, INSERT ON `data1`.*' in result.stdout"
+          - "'GRANT SELECT, DELETE ON `data2`.*' in result.stdout"
+
+    - name: Append privileges that are a subset of the current privileges, which should be a no-op
+      mysql_user:
+        <<: *mysql_params
+        name: '{{ user_name_4 }}'
+        password: '{{ user_password_4 }}'
+        priv: 'data1.*:SELECT/data2.*:SELECT'
```
Andersson007:
```suggestion
        priv: 'data1.*:SELECT/data2.*:SELECT'
```
@steveteahan thanks for this!
Could you please also add a test case expanding privileges for one of them after this one, e.g.:
`priv: 'data1.*:DELETE/data2.*:SELECT'`, then check

These changes look logical to me but i myself am not using the module.
While you're adding the case I'll email our WG leader to look at the PR

bmalynovytch:
Code looks good to me 👍

bmalynovytch (approved):

tests/integration/targets/test_mysql_user/tasks/test_priv_append_no_change.yml
steveteahan: Added the requested test.

Andersson007 (approved):

Andersson007:
@steveteahan thanks for fixing this!
@bmalynovytch thanks for reviewing!
Good job, folks!

@steveteahan if you like to do more, we have several issues / PRs to fix / review:) Just a FYI

Andersson007:
FYI: i'll release the next patch version of the collection after https://github.com/ansible-collections/community.mysql/pull/71 is merged

Issue #91, opened 2021-01-14
Supported database versions and integration testing


steveteahan:
##### SUMMARY
I'm creating this issue hoping to start a discussion about which database server/versions that this collection supports and how that affects integration testing.  Based on the code, it seems that the intent is to support MySQL <5.7, MySQL 5.7, MySQL 8.0, and MariaDB.  The integration tests only execute on MySQL 5.7 and 8.0, however, which limits the guarantees of compatibility for all other servers and versions.

I had some follow-up questions that the maintainers may or may not find are worth discussing:

- Would it be worth explicitly documenting the database servers and versions that the collection supports?
- If the above makes sense, should there be a standard for dropping support?  For instance, support will be dropped for a particular version after the end of support date?  A 3-6 month buffer could seem reasonable too?  This could also provide opportunities to remove old code paths.
- Should all the supported versions be included in integration tests?  Or are there specific limitations that are preventing doing so?

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
N/A

##### ADDITIONAL INFORMATION

It is worth mentioning that it looks like end of support for MySQL 5.6 appears to be quickly approaching, so I can see that it might not be worth including that version in integration tests now.  This isn't the case for MariaDB, however.

Some examples where coverage is being missed:

- https://codecov.io/gh/ansible-collections/community.mysql/src/main/plugins/modules/mysql_user.py#L362
- https://codecov.io/gh/ansible-collections/community.mysql/src/main/plugins/modules/mysql_user.py#L504
- https://codecov.io/gh/ansible-collections/community.mysql/src/main/plugins/modules/mysql_user.py#L600
- https://codecov.io/gh/ansible-collections/community.mysql/src/main/plugins/modules/mysql_user.py#L719
- https://codecov.io/gh/ansible-collections/community.mysql/src/main/plugins/modules/mysql_user.py#L911

Jorge-Rodriguez:
@steveteahan I very much she with you, and it's true that we're not testing against MariaDB and we should.
The reason for this is that, AFAIK we do not have a MariaDB container setup to test against. Currently the tests run against a Percona cluster.

If someone could put the work into setting up such a container (I believe there work would go into the setup.yml file of the integration tests) we could start testing against MariaDB.

steveteahan:
[...]

PR #92, opened 2021-01-16
mysql_user - added queries to module output and set no_log option on …

### LINKED ISSUE #75: mysql_user: Adding statements output
##### SUMMARY
Adding a statements list to the output of mysql_user could make the module more transparent and easier to debug.  Ideally, users would be able to run in check mode and see exactly which statements would be executed to get to the desired state.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
mysql_user

##### ADDITIONAL INFORMATION
The mysql_user module currently returns **changed**, **user**, and **msg**.  This proposal would add **statements** to include the SQL statements associated with the actual user or permission changes.  An example of the proposed output may be seen below:

```
    "statements": [
        "CREATE USER 'db_user2'@'localhost' IDENTIFIED WITH mysql_native_password AS '[redacted]'",
        "GRANT EXECUTE ON FUNCTION `foo`.`function` TO 'db_user2'@'localhost'",
        "GRANT SELECT ON `foo`.* TO 'db_user2'@'localhost'",
        "GRANT EXECUTE ON PROCEDURE `bar`.`procedure` TO 'db_user2'@'localhost'",
        "GRANT USAGE ON *.* TO 'db_user2'@'localhost'"
    ],
```

An initial version (not ready for PR) can be seen here: https://github.com/steveteahan/community.mysql/commit/245f5f77a57b236fc926ad533a2a8fc4b7a26a0f

Providing the output of the SQL statements is something that the postgresql collection is doing today:

* https://github.com/ansible-collections/community.postgresql/blob/main/plugins/modules/postgresql_privs.py#L1168
* https://github.com/ansible-collections/community.postgresql/blob/main/plugins/modules/postgresql_user.py#L988

In that collection, they call it **queries** which I think is acceptable is well, it just seemed the **statements** was possibly more accurate.

Looking for some initial feedback about whether the maintainers are open to this change, or if it was already considered and avoided.  Some other general thoughts about this:

* Most of the functions in the module return early in check_mode, before the query is ever built, so additional refactoring would be required to make this work in check mode.
* Check mode is arguably where this functionality would be most useful, so I'll take a look at the feasibility of moving the check mode conditionals around `cursor.execute`, but it's possible it wouldn't be that easy.

steveteahan:
…plugin_hash_string and plugin_auth_string (#75)

##### SUMMARY
Fixes #75 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
mysql_user

##### ADDITIONAL INFORMATION
* There will probably be a couple of uncovered areas which would be resolved either by dropped MySQL <5.7 support or adding tests for MySQL 5.6.  Tracking a discussion of that topic in #91.
* In `user_mod` I skipped adding the queries in an exception case because it currently doesn't have test coverage and I couldn't reproduce the issue in either MySQL 5.7 or 8.0.  The code in question is here: https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_user.py#L601

### DIFF
--- a/changelogs/fragments/75-mysql_user_queries_output.yml
+++ b/changelogs/fragments/75-mysql_user_queries_output.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- mysql_user - added queries to module output and set no_log option on plugin_hash_string and plugin_auth_string (https://github.com/ansible-collections/community.mysql/issues/75).
+- mysql_user - added queries to module output and set `no_log` flag on `plugin_hash_string` and `plugin_auth_string` (https://github.com/ansible-collections/community.mysql/issues/75).
--- a/changelogs/fragments/75-mysql_user_queries_output.yml
+++ b/changelogs/fragments/75-mysql_user_queries_output.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- mysql_user - added queries to module output and set no_log option on plugin_hash_string and plugin_auth_string (https://github.com/ansible-collections/community.mysql/issues/75).
--- a/plugins/modules/mysql_user.py
+++ b/plugins/modules/mysql_user.py
@@ -300,7 +300,7 @@ queries:
   returned: if executed successfully
   type: list
   sample: ["CREATE USER 'db_user2'@'localhost' IDENTIFIED WITH mysql_native_password AS '********'"]
-  version_added: '1.1.0'
+  version_added: '1.3.0'
 '''
 
 
--- a/plugins/modules/mysql_user.py
+++ b/plugins/modules/mysql_user.py
@@ -294,7 +294,15 @@ EXAMPLES = r'''
 # password=n<_665{vS43y
 '''
 
-RETURN = '''#'''
+RETURN = r'''
+queries:
+  description: List of executed queries which modified DB's state.
+  returned: if executed successfully
+  type: list
+  sample: ["CREATE USER 'db_user2'@'localhost' IDENTIFIED WITH mysql_native_password AS '********'"]
+  version_added: '1.1.0'
+'''
+
 
 import re
 import string
@@ -347,9 +355,24 @@ VALID_PRIVS = frozenset(('CREATE', 'DROP', 'GRANT', 'GRANT OPTION',
 class InvalidPrivsError(Exception):
     pass
 
+
+# Same as what is used by Ansible no_log parameters for consistency.
+REDACTED_AUTH_STRING = "*" * 8
+
+_executed_queries = []
+
 # ===========================================
 # MySQL module specific support methods.
 #
+
+
+def log_query(cursor, query, args=None):
+    """
+    Logs the query including the args provided.
+
+    Note: check that all args handle credentials properly (parameters should have no_log set or other mitigations).
+    """
+    _executed_queries.append(cursor.mogrify(query, args))
 
 
 # User Authentication Management changed in MySQL 5.7 and MariaDB 10.2.0
@@ -490,6 +513,8 @@ def user_add(cursor, user, host, host_all, password, encrypted,
 
     mogrify = do_not_mogrify_requires if old_user_mgmt else mogrify_requires
 
+    sanitized_query_with_args = ""
+
     if password and encrypted:
         if supports_identified_by_password(cursor):
             query_with_args = "CREATE USER %s@%s IDENTIFIED BY PASSWORD %s", (user, host, password)
@@ -502,6 +527,14 @@ def user_add(cursor, user, host, host_all, password, encrypted,
             cursor.execute("SELECT CONCAT('*', UCASE(SHA1(UNHEX(SHA1(%s)))))", (password,))
             encrypted_password = cursor.fetchone()[0]
             query_with_args = "CREATE USER %s@%s IDENTIFIED WITH mysql_native_password AS %s", (user, host, encrypted_password)
+
+            # encrypted_password above is not a module argument and therefore cannot use no_log to prevent logging. This
+            # is a workaround for that. Using 'BY' instead of 'AS' in the statement appears to allow passing in
+            # plaintext values for MySQL 5.7 and 8.0. It is not clear whether this was avoided for a specific reason.
+            # This special case can be removed if the change described above is implemented in the future.
+
+            sanitized_query_with_args = "CREATE USER %s@%s IDENTIFIED WITH mysql_native_password AS %s",\
+                                        (user, host, REDACTED_AUTH_STRING)
     elif plugin and plugin_hash_string:
         query_with_args = "CREATE USER %s@%s IDENTIFIED WITH %s AS %s", (user, host, plugin, plugin_hash_string)
     elif plugin and plugin_auth_string:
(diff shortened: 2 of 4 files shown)


steveteahan:
Looks like mysqlclient doesn't have a `mogrify` method:

```
AttributeError: 'Cursor' object has no attribute 'mogrify'
```

I'll have to take a look at what the other options are.  It seems like in other areas of the project the cursor [private variables](https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_query.py#L230) are being used which isn't ideal.  In others, it seems to be simple string replacement which from my understanding won't work for all cases.

steveteahan:
I created an issue for the mysqlclient project to see if adding mogrify would be considered: https://github.com/PyMySQL/mysqlclient/issues/476

That seems like one of the better solutions that I can think of at the moment.  The worst-case could be to only log the queries output if the MySQL connector in use supports `mogrify`?

steveteahan:
I have a PR up for review for the mysqlclient change.  Even when that is implemented it will only work for the latest version when released, so I'm thinking that I'll still check the cursor for `mogrify()` and if it doesn't exist simply add a warning to the output.

changelogs/fragments/75-mysql_user_queries_output.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- mysql_user - added queries to module output and set no_log option on plugin_hash_string and plugin_auth_string (https://github.com/ansible-collections/community.mysql/issues/75).
```
Andersson007:
```suggestion
- mysql_user - added queries to module output and set `no_log` flag on `plugin_hash_string` and `plugin_auth_string` (https://github.com/ansible-collections/community.mysql/issues/75).
```

plugins/modules/mysql_user.py
```diff
@@ -294,7 +294,15 @@
 # password=n<_665{vS43y
 '''
 
-RETURN = '''#'''
+RETURN = r'''
+queries:
+  description: List of executed queries which modified DB's state.
```
Andersson007, line 299:
```suggestion
  description: List of executed queries which modified a state of a database.
```
We shouldn't use abbreviations if possible

plugins/modules/mysql_user.py
```diff
@@ -294,7 +294,15 @@
...
+  returned: if executed successfully
+  type: list
+  sample: ["CREATE USER 'db_user2'@'localhost' IDENTIFIED WITH mysql_native_password AS '********'"]
+  version_added: '1.1.0'
```
```suggestion
  version_added: '1.3.0'
```
I released 1.2.0 two days ago, so it should be the next minor release

Andersson007:
@steveteahan thanks for working on this!
I'm not sure about the warning - it could annoy users who don't need to see any returned values.
Maybe:
1. check if `mogrify()` is supported once in the `main` by catching the exception, define a boolean variable to indicate that
2. if not supported, append to `_executed_queries` something like meaning "You MySQL driver doesn't support ... Please use ..."
3. pass a boolean flag to `log_query()` function and just return with no action OR add a condition not to invoke the function at all

Just a suggestion, I don't feel fresh now:)

Also the codecov-related tests should be green.

Andersson007:
BTW i don't see a good explanation in the module's documentation of what `plugin_hash_string` and `plugin_auth_string` are about. Even only one example is provided.
Does it really make sense to hide them in output?

plugins/modules/mysql_user.py
steveteahan: I copied that from `mysql_replication.py` for consistency, but agree that it could be better.  I'll fix it.

steveteahan:
> @steveteahan thanks for working on this!
> I'm not sure about the warning - it could annoy users who don't need to see any returned values.
> Maybe:
> 
>     1. check if `mogrify()` is supported once in the `main` by catching the exception, define a boolean variable to indicate that
> 
>     2. if not supported, append to `_executed_queries` something like meaning "You MySQL driver doesn't support ... Please use ..."
> 
>     3. pass a boolean flag to `log_query()` function and just return with no action OR add a condition not to invoke the function at all

Moving away from a warning seems reasonable to me if your experience in the community has been that it is more of annoyance than doing any good.  I can do something like what is suggested, but maybe leave the appending to `queries` until the very end where I'm currently calling `module.warn`.  That way the message would only show up once instead of many times.

> 
> Also the codecov-related tests should be green.

One of the issues is not testing MySQL 5.6 which I tried to start a discussion about in #91.  The other part was never covered, but I'll try to get a test together for it.

> BTW i don't see a good explanation in the module's documentation of what `plugin_hash_string` and `plugin_auth_string` are about. Even only one example is provided.
> Does it really make sense to hide them in output?

`plugin_auth_string` is a plaintext password and `plugin_hash_string` is in the format that the database expects it to be in (generally hashed).  This is covered in a bit more detail [here](https://dev.mysql.com/doc/refman/8.0/en/create-user.html#create-user-authentication).  From that perspective, I think it definitely makes sense to hide them from output.

plugins/modules/mysql_user.py
```diff
@@ -592,11 +633,15 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
                 if module.check_mode:
                     return (True, msg)
                 if old_user_mgmt:
-                    cursor.execute("SET PASSWORD FOR %s@%s = %s", (user, host, encrypted_password))
+                    query = "SET PASSWORD FOR %s@%s = %s"
```
steveteahan, line 636: This is the example of the missing coverage for MySQL 5.6 that I mentioned above.

I did a bit of research and enabling integration tests for MySQL 5.6 is not going to be trivial (see #91).  `old_user_mgmt` handles cases where MariaDB < 10.2 and MySQL < 5.7.  The only version in that set that is still supported is MySQL 5.6 (for another two weeks).  I don't see it being worthwhile spending any time on this.  As per the discussion is #91 it might be worthwhile to consider dropping this support and removing the code in the near future.

Unless there are other suggestions, I can go ahead and remove this change.  The downside is that until we remove MariaDB < 10.2 and MySQL < 5.7 code paths, there will be queries that can modify the database state, but will not be logged via the module.

Andersson007:
> I can do something like what is suggested, but maybe leave the appending to queries until the very end where I'm currently calling module.warn.

I meant it'll be append once, then will be skipped. But feel free to implement that as you like, will see

> plugin_auth_string is a plaintext password and plugin_hash_string is in the format that the database expects it to be in (generally hashed). This is covered in a bit more detail here. From that perspective, I think it definitely makes sense to hide them from output.

If so, sounds sensible

steveteahan:
The `mogrify` change hasn’t been merged into mysqlclient yet. I also don’t have an immediate need for this change any longer. I still think it’s a change that adds value. I’ll leave this PR open and leave it up to the maintainers to decide if it should stay open for someone else to push over the line once mysqlclient is ready, or if this should just be closed.

Thanks!

Jorge-Rodriguez:
@Andersson007 @steveteahan The codebase has changed dramatically since opening this PR. At this point I don't know if rebasing is going to be any easier than replicating the work on top of the current head.

Andersson007:
> @Andersson007 @steveteahan The codebase has changed dramatically since opening this PR. At this point I don't know if rebasing is going to be any easier than replicating the work on top of the current head.

Agreed

steveteahan:
Closing this PR based on the feedback above.

Andersson007:
@steveteahan thank you!

PR #95, opened 2021-02-09
Update mysql_user.py

tvlooy:
cleanup:
- mysql_user - the "REPLICATION SLAVE" is duplicate in the list. Spotted by @tomcan

### DIFF
--- a/plugins/modules/mysql_user.py
+++ b/plugins/modules/mysql_user.py
@@ -338,7 +338,6 @@ VALID_PRIVS = frozenset(('CREATE', 'DROP', 'GRANT', 'GRANT OPTION',
                          'CONNECTION ADMIN',
                          'READ_ONLY ADMIN',
                          'REPLICATION MASTER ADMIN',
-                         'REPLICATION SLAVE',
                          'REPLICATION SLAVE ADMIN',
                          'SET USER',
                          'SHOW_ROUTINE',))


Andersson007:
@tvlooy thanks for the PR @TomCan thanks for pointing it out!
@tvlooy could you please
1) change the commint message to something like:
```mysql_user: remove duplicate of REPLICATION SLAVE entry from valid privs frozenset```

2) this is a bugfix, so please add a changelog fragment https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment
with content like:
```
bugfixes:
- mysql_user - the module crashes when trying to grant ``REPLICATION SLAVE`` priviledge (https://github.com/ansible-collections/community.mysql/pull/95).
```
or something similar

Andersson007:
My comment above is relevant, of course, if you're sure that this fixes the problem. If it doesn't and you just noticed the duplicate, change the commit message and a changelog fragment entry correcpondingly. Thanks!

tvlooy:
@Andersson007 thanks for the review. It's not a bugfix in the sense that it does not fix a bug. Just something we noticed when digging into an issue (on our end, that has been solved in the meanwhile). So marked as "cleanup"

Andersson007:
@tvlooy @TomCan thanks!

PR #111, opened 2021-03-09
[DRAFT] mysql new lookup plugin

Andersson007:
##### SUMMARY
mysql new lookup plugin to fetch data from mysql tables

It returns a list of dictionaries `{col_name: value}`, for example:
```
[
            {
                "id": 3,
                "story": "first"
            },
            {
                "id": 4,
                "story": "second"
            }
]
```

Several important questions to discuss:

1. I've never written plugins before, decided to try BUT do we need such a plugin? Would it be really useful for users? Yes, we have the `mysql_query` module but it's a module, not a plugin. In the plugin form, it can be potentially convenient to use. Don't hesitate to criticize this. If it's not worth enough, it's better not to add this.

2. If it's worth.  Are the interfaces (e.g., the options, return values, etc.) good? Maybe we should return a list of tuples instead of dictionaries (I personally prefer dictionaries)? Maybe we should add more options (however we shouldn't overwhelm it, otherwise the plugin's invocation can be difficult to read)?

3. Maybe we should fetch and return data drastically different way?

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
plugins/lookup/mysql.py

### DIFF
--- a/plugins/lookup/mysql.py
+++ b/plugins/lookup/mysql.py
@@ -1,0 +1,194 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2021, Andrew Klychkov (@Andersson007) <aaklychkov@mail.ru>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+name: mysql
+
+short_description: Fetch data from MySQL table
+
+description:
+  - Fetch data from MySQL table.
+  - For complex queries use M(community.mysql.mysql_query) module.
+
+version_added: '1.3.0'
+
+options:
+  _terms:
+    description:
+      - Tables to fetch data from.
+      - "You can use colons to pass optional column list, for example, 'mytable:col1,col2'."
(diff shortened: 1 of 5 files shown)


Andersson007:
CC @Jorge-Rodriguez @bmalynovytch 
@felixfontein would be very nice to hear your opinion on the questions from the description

Andersson007:
from @felixfontein in IRC
```
andersson007_: 1) the plugin reads all rows form the tables - that doesn't seem very useful to me in most cases; 2) you should allow to set all the config options with vars (for how to do that: https://github.com/ansible-collections/community.sops/blob/main/plugins/lookup/sops.py#L78 https://github.com/ansible-collections/community.sops/blob/main/plugins/doc_fragments/sops.py#L69)
9:50 AM config options = 'general connection infos'
9:51 AM (also you don't escape all the things you put into the query :) )
9:51 AM another idea: why not allow to specify columns as a list option (columns=['a', 'b']) instead of parsing the string?
```

Andersson007:
I think if there's no demand for this plugin, and, moreover, we have mysql_query module, I should close the PR. Will see, maybe later if there's an issue for that.
@felixfontein thanks for reviewing! If once I decide to write a plugin or will review others plugins, I'll look at your comment here first:)

PR #116, opened 2021-03-12
change deprecated parameter pw and db

ziegenberg:
##### SUMMARY

Parameter `db` and `passed` are deprecated in PyMySQL since v1.0.0 and are kept as an alias for `database` and `password`. Warnings will be activated around 2022 (for details see https://github.com/PyMySQL/PyMySQL/issues/939).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

module_utils

### DIFF
--- a/changelogs/fragments/116-change_deprecated_connection_ parameters.yml
+++ b/changelogs/fragments/116-change_deprecated_connection_ parameters.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- mysql module utils - change deprecated connection parameters ``passwd`` and ``db`` to ``password`` and ``database`` (https://github.com/ansible-collections/community.mysql/pull/116).
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -79,7 +79,7 @@ def mysql_connect(module, login_user=None, login_password=None, config_file='', 
     if login_user is not None:
         config['user'] = login_user
     if login_password is not None:
-        config['passwd'] = login_password
+        config['password'] = login_password
     if ssl_cert is not None:
         config['ssl']['cert'] = ssl_cert
     if ssl_key is not None:
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -87,7 +87,7 @@ def mysql_connect(module, login_user=None, login_password=None, config_file='', 
     if ssl_ca is not None:
         config['ssl']['ca'] = ssl_ca
     if db is not None:
-        config['db'] = db
+        config['database'] = db
     if connect_timeout is not None:
         config['connect_timeout'] = connect_timeout
     if check_hostname is not None:


Andersson007:
@ziegenberg hi, thanks for the PR!
Please add a changelog fragment with `minor_changes:` see https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment (please follow the format), for example:
```
minor_changes:
- mysql module utils - change deprecated connection parameters ``passwd`` and ``db`` to ``password`` and ``database`` (https://github.com/ansible-collections/community.mysql/pull/116).
```

https://dev.mysql.com/doc/connector-python/en/connector-python-connectargs.html also doesn't recommend to use `db` and `passwd`

@bmalynovytch @Jorge-Rodriguez what do you think? As far as i know we test against both pymysql and mysqldb and the tests seem to be green.
If you're ok, feel free to to push the big green button

ziegenberg:
@Andersson007 changelog fragment added.

Andersson007:
@ziegenberg thanks! BTW don't worry about codecov

Andersson007:
Let's wait for the folks opinion

Jorge-Rodriguez:
@ziegenberg I'm worried about the backwards compatibility of this change. Do you have the version numbers when these options were deprecated?

ziegenberg:
It was introduced into PyMySQL with this PR  on Jun 6, 2014: https://github.com/PyMySQL/PyMySQL/pull/247
If I saw this correctly, it's available at least since version 0.6.3.

Jorge-Rodriguez:
@ziegenberg I'm afraid this change breaks the compatibility with MySQLdb connector that still uses the parameter names deprecated in pymysql, as seen [here](https://mysqlclient.readthedocs.io/user_guide.html#functions-and-attributes)

But then, the tests pass, so it must be an issue with MySQLdb's documentation, and since we're testing all the connector versions that we support, I suppose it's all good.

Andersson007:
@ziegenberg thanks for the PR!
@Jorge-Rodriguez thanks for reviewing and merging!

ziegenberg:
@Jorge-Rodriguez in [MySQLdb](https://github.com/PyMySQL/mysqlclient/blob/master/MySQLdb/connections.py) they also have `password` and `passwd` as an alias, so I guess updating the docu, as you said, seems the right approach.

ziegenberg:
Ah, they already have updated the docs in their main branch. See this commit https://github.com/PyMySQL/mysqlclient/commit/fa25358d0f171bd8a63729c5a8d76528f4ae74e9

tschoonj:
I am seeing the following error due to this change:

```
failed: [vm-ip] (item=rfi-stfc-guacamole-vm.novalocal) => {"ansible_loop_var": "item", "changed": false, "item": "rfi-stfc-guacamole-vm.novalocal", "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: 'database' is an invalid keyword argument for this function"}
```

The remote is using `MySQL-python-1.2.5-1.el7.x86_64`.

tschoonj:
Switching to PyMySQL fixed things for me, but looks like backwards compatibility was broken with MySQLdb.

ziegenberg:
I'm not sure if MySQL-python-1.2.5 is a supported version of this collection? It was released in 2014.

Jorge-Rodriguez:
As of now, only MySQLdb version 2 is supported.

tschoonj:
Ok that's fine, but it might be wise to update the docs accordingly.

Jorge-Rodriguez:
You're very correct about that. And it's actually planned, along with the list of supported server versions.

awegrzyn:
Hi,
I've spent a bit of time today to debug an issue with our deployment as we've been getting `Exception message: 'database' is an invalid keyword argument for this function"`.
Would it be possible to deprecate `MySQL-python-1.2.5` more gracefully?
Docs still says: `Requires the PyMySQL (Python 2.7 and Python 3.X) or MySQL-python`, and on C7 `MySQL-python` brings 1.2.5...
Thanks.

Jorge-Rodriguez:
> and on C7 `MySQL-python` brings 1.2.5...

Is that CentOS 7?

Andersson007:
https://github.com/ansible-collections/community.mysql/issues/151
1) maybe we could just quickly revert this and release? The change doesn't bring a lot in comparison with the consequences

Andersson007:
or 2) quick patch like "if this error occurs, use `db`" ?

Jorge-Rodriguez:
Let's revert and consider a more seamless approach to reimplement it

Jorge-Rodriguez:
@ziegenberg could you check what was the version of MySQLdb that introduced the new database and password keywords?

ziegenberg:
A quick check brought up that it might be v1.3.8 in 2016. I'll check thoroughly when I'm back at my desk. Currently only on a mobile.

Andersson007:
Reverted with https://github.com/ansible-collections/community.mysql/pull/153. will try to release today.
I think the best approach is to check driver versions and behave correspondingly.
See the surrounding code in the mysql_connect function - there's an example of the check.

Thanks everyone for collaborating!

Andersson007:
Released:) Who wants to handle it gracefully now, welcome:)

Andersson007:
Available on galaxy https://galaxy.ansible.com/community/mysql. How to install from galaxy, refer to https://galaxy.ansible.com/community/mysql, or wait for the next Ansible package release.

Andersson007:
@ziegenberg hi, would you like to continue working on this?

ziegenberg:
I can give it another try. But first, I want to say sorry for the disruption my contribution has caused.

Before implementing this, I guess we need to sort out the supported versions of the connectors. And fix the CI accordingly. I think this should go into a separate PR beforehand.

Andersson007:
> I can give it another try. But first, I want to say sorry for the disruption my contribution has caused.

Don't worry about it much, this happens to all of us from time to time (for example, I broke and am breaking stuff a lot but it gives valuable experience). It's normal. Now we know that we should be careful about driver versions even if they seem to be absolutely obsolete. 

> Before implementing this, I guess we need to sort out the supported versions of the connectors. And fix the CI accordingly. I think this should go into a separate PR beforehand.

Sounds very sensible.
I think the related files are stored in `.github/workflows/` and `tests/integration/targets/setup_mysql/`. If i missed something @Jorge-Rodriguez can correct me.

@ziegenberg feel free to create related issues / PRs and ask if there are any questions on your mind.

Thanks!

Jorge-Rodriguez:
@ziegenberg I'm working on a PR tackling CI versions, you can keep an eye on #163

ziegenberg:
I've created #176 to track the discussion of changing deprecated parameter pw and db.

PR #117, opened 2021-03-13
Mark mysql_replication's master_ssl_key option as no_log=False

felixfontein:
##### SUMMARY
Should fix CI (sanity tests currently fail because of that).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
mysql_replication

### DIFF
--- a/plugins/modules/mysql_replication.py
+++ b/plugins/modules/mysql_replication.py
@@ -435,7 +435,7 @@ def main():
         master_ssl_ca=dict(type='str'),
         master_ssl_capath=dict(type='str'),
         master_ssl_cert=dict(type='str'),
-        master_ssl_key=dict(type='str'),
+        master_ssl_key=dict(type='str', no_log=False),
         master_ssl_cipher=dict(type='str'),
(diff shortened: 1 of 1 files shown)


felixfontein:
I didn't add a changelog fragment since this should not be user-facing. If you'd like to have one, please tell me :)

CC @Andersson007

felixfontein:
Also please someone verify that this is indeed not a secret. I didn't find any documentation on such a MySQL variable, but similar ones seemed to be filenames (and not the secret key itself), so I *assume* it's the same here.

If my assumption is wrong, this should be a security fix and `no_log` should be set to `True`.

Andersson007:
@felixfontein yeah, it's a path https://dev.mysql.com/doc/refman/8.0/en/replication-solutions-encrypted-connections.html,
thanks for the fix!:)

felixfontein:
@Andersson007 thanks! It might make sense to link to these docs from the options then ;-)

Andersson007:
@felixfontein good spot, will think, thanks:)

PR #147, opened 2021-04-15
mysql_replication: add deprecation warning about future replacement of Is_Slave and Is_Master return values, add alternatives

Andersson007:
##### SUMMARY
mysql_replication:
1) add deprecation warnings about future replacement of `Is_Slave` and `Is_Master` return values with `Is_Primary` and `Is_Replica`
2) add the alternative return values

PR #149, opened 2021-04-16
[stable-2] mysql_replication: add deprecation warning about future replacement of Is_Slave and Is_Master return values, add alternatives (#147)

Andersson007:
(cherry picked from commit 853db5a2a415cb2b9dbcea0c0c65aec64e9a2f82)

##### SUMMARY
Backport of https://github.com/ansible-collections/community.mysql/pull/147


Andersson007:
cc @Jorge-Rodriguez @bmalynovytch
No need to review, just an example of the first backport to `stable-2` which we should do for our and others' PRs we merge (except breaking changes, they will stay in `main` only).

There can be also cases, when a somebody else's PR contains both breaking and not breaking changes. In this case it would be
better to ask authors to split such PRs, otherwise we will have to extract non-breaking changes manually and backport which is not good:)

I'll write a short backport_guide.rst

If any questions, feel free to ask.

Andersson007:
and yesterday i forgot to create `stable-1` - created now. We should backport only bugfixes to `stable-1`.

I will try to use the backport bot to simplify the process.

PR #189, opened 2021-06-24
mysql_role: new module

### LINKED ISSUE #50: Create ROLES, assign them and SET default role
##### SUMMARY
Implement ROLE feature which is part of MySQL 8 / MariaDB 10
https://mariadb.com/kb/en/roles_overview/
https://dev.mysql.com/doc/refman/8.0/en/create-role.html
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
mysql_user

##### ADDITIONAL INFORMATION
I didn't find any way to create a ROLE (collection of GRANTS) which can be assigned to a USER. 
Those ROLES doesn't have a host and can make permission handling easier when you have many users. 
You also can set a default role of a user so it is transparent for the application.

Did I miss anything?
Regards
Markus

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.mysql/issues/50

mysql_role: new module

Not sure if I'm able to finish this (I'll put it explicitly if not). Hope it'll help anyway (i.e. to define an interface, etc.)

IMPORTANT
- I moved most of the code from the `mysql_user` module to the `user` library.
-  in the `user_mod` function I added a couple of conditions related to roles, look for the `role` variable to find those  places.
-  I added a lot of `maria_role` across `plugins/module_utils/user.py` (review it carefully).
-  the module uses the `user_mode` function and it always reports `changed=True` when `append_privs=yes`. I wonder if it's expected behavior for the `mysql_user` module?  (UPDATE: yes, this is the same there, checked, I'll create an issue to fix the bug alter this gets merged to avoid conflicts)
- in a several moths after this released, we can move the Role class to a library and use in the `mysql_user` module to add a user to roles via a new option (called, say, `roles:`).
- integration tests were added and they cover (I hope) everything available on MySQL
- as we don't have MariaDB in our CI, i tested things manually locally on MariaDB. So if anyone tests it and confirms that everything is fine (or not), it would be great.
- I covered a couple of classes and functions with unit tests as well. As i spent more than a week to write this, i have no wish to cover the rest at least now. As the things are covered with integration tests, it's not crucial

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
mysql_role


Andersson007:
recheck

Andersson007:
cc @Jorge-Rodriguez @bmalynovytch @RealKelsar @Markus- @camillehuot ready for review

(there's a codecove error, we can ignore)

(if anyone tests the `admin` option manually with MariaDB, it would be great)

Andersson007:
I tested it manually with MariaDB... it was painful. Found differences. Seems to be fixed, though it can be a wrong feeling.
Would be nice if anybody else test it manually with it as well.
Ready for review.

plugins/modules/mysql_role.py
```diff
@@ -0,0 +1,784 @@
...
+    def __role_exists(self):
+        """Check if a role exists.
+
+        Returns:
+            bool: True if the role exists, False if it does not.
+        """
+        if not self.is_mariadb:
+            query = 'SELECT count(*) FROM mysql.user WHERE user = %s AND host = %s', (self.name, self.host)
+        else:
+            query = "SELECT count(*) FROM mysql.user WHERE user = %s AND host = ''", (self.name)
```
camillehuot: What about filtering on the `is_role` column?
```
SELECT count(*) FROM mysql.user WHERE user = %s AND is_role = 'Y'
```

plugins/modules/mysql_role.py
```diff
@@ -0,0 +1,784 @@
...
+            return False
+
+        changed = False
+        for user in users:
+            if user not in self.members:
+                if check_mode:
+                    return True
+
+                if not self.is_mariadb:
+                    self.cursor.execute('GRANT %s@%s TO %s@%s', (self.name, self.host, user[0], user[1]))
```
camillehuot: I guess you can avoid this `if` (and maybe others) by using `self.full_name` to refer to the role name

plugins/modules/mysql_role.py
```diff
@@ -0,0 +1,784 @@
...
+            if detach_members:
+                changed = self.remove_members(users, check_mode=check_mode)
+
+            else:
+                changed = self.add_members(users, check_mode=check_mode,
+                                           append_members=append_members,
+                                           set_default_role_all=set_default_role_all)
+
+        if privs:
+            changed, msg = user_mod(self.cursor, self.name, self.host,
```
camillehuot, line 861: It looks like when both `users` and `privs` are set, you override the value set by the previous block for `changed`.
Ex. when you add a member but there is no change in the role privileges, you would return `False` instead of `True`.

plugins/modules/mysql_role.py
Andersson007: Yes, we can but it's better to pass variables through parameters whenever possible.

plugins/modules/mysql_role.py
Andersson007: very good spot, thanks! fixed

plugins/modules/mysql_role.py
Andersson007: Done, thanks!

Andersson007:
it's turning out to be harder to handle differences between MySQL and MariaDB, so I marked this as a draft again.

Andersson007:
Ready for review (I updated the description, there are important information, please take a look at it first)

Andersson007:
I updated README. Hey, folks, Friday is a good day for reviewing

Andersson007:
I hope we'll get more reviews here. I'm gonna merge this anyway in a week or two and release a prerelease so that this won't be included in Ansible package but interested users will be able to download it via `ansible-galaxy` easily and test.

Andersson007:
community.mysql 2.2.0-a1 has been released and available to install via galaxy or directly from https://galaxy.ansible.com/community/mysql. It contains the `mysql_role` new module. `Pay attention that this is a pre-release and is not considered ready for production`.
Any feedback would be much appreciated.
After confirming that everything is find with the module (and with `mysql_user` module as well because both the modules share some code), we can release 2.2.0.

Issue #238, opened 2021-10-19
CI: Add MariaDB support (i.e. let integration tests run using MariaDB as a server)


Andersson007:
##### SUMMARY

As there are issues specific to MariaDB reported and our fixes / features can affect MariaDB users, we need to have MariaDB in CI

Related files:
- .github/workflows/ansible-test-plugins.yml
- tests/integration/targets/setup_mysql/*

There's also currently not used role https://github.com/ansible-collections/community.mysql/tree/main/tests/integration/old_mariadb_replication that were used when the collection's content lived in ansible/ansible (might be helpful)

##### ISSUE TYPE
- Feature Idea

PR #243, opened 2021-10-30
Get rid of privs comparison

### LINKED ISSUE #234: mysql_user: get rid of privs comparison
##### SUMMARY
(this is a thing to discuss)

We should analyze possibility to get rid of privs comparison entirely and use just a mechanisms like "try except": let's let the server decide which privs are supported and throw the exception itself.

A **solution** can conceptually look like the following:
- Remove priv comparison
- Add exception handling to catch an exception thrown by the DB (maybe there's something currently present)
- The handler (try except + parsing the error for related keywords) will notify the user and exit gracefully (i.e. without ugly tracebacks, etc.)

However this is a breaking change and we'll require a major release in the future.

Possible scenarios:
**Option 1 (Most votes are for this variant)**:
- Implement the above (or another solution) now
- Test it thoroughly with MariaDB and MySQL
- Merge it after 3.0.0 release (within 2 months)
- Notify users via the changelog that such changes will be released in the major release (now it'll be 4.0.0 in 6-8 months or later)
- Release them finally

**Option 2 (not breaking but downside is one more argument)**:
- Implement the solution above without removing the priv comparison (i.e. in parallel)
- Add a new bool argument called, say, `check_privs` with default `yes`)
- If the option is `no` - comparison will be skipped
- Test it with MySQL, MariaDB
- Release it right now and forget - users will be able to switch back and forth if they need

**Option 3**:
- Same as Option 2 but with a warning to users that the `check_privs` will get `no` as its default in 4.0.0
- Plus announcing this via the changelog
- Finally do it in 4.0.0

**Option 4**:
- ...

##### CONTEXT

- Before 2.3.0 all supported privs were hardcoded in the `VALID_PRIVS` frozenset.
- Since 2.3.0 all supported privs are fetched from the server dynamically with the `SHOW PRIVILEGES` command.
- The issue is that some privs were deprecated, replaced and became aliases and they are not present in the `SHOW PRIVILEGES` output but they were present in the `VALID_PRIVS` frozenset.
- Removing `VALID_PRIVS` brought breaking changes.
- Relates to https://github.com/ansible-collections/community.mysql/pull/233 fixed the breaking change - returned `VALID_PRIVS` and now the module merges it in a frozenset with `show_privs` (fetched from server dynamically) and `EXTRA_PRIVS`.

Relates to https://github.com/ansible-collections/community.mysql/pull/233
Relates to https://github.com/ansible-collections/community.mysql/pull/217

rsicart:
##### SUMMARY
Fixes #234 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
user, mysql_user, mysql_role

##### ADDITIONAL INFORMATION
See #234 for more details.

### DIFF
--- a/tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
+++ b/tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
@@ -112,7 +112,7 @@
       assert:
         that:
           - "result.changed == false"
-          - "'Illegal privilege' in result.msg or 'You have an error in your SQL syntax;' in result.msg"
+          - "'Error granting privileges' in result.msg"
       when: enable_check_mode == 'no'
 
     ##########
--- a/tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
+++ b/tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
@@ -96,6 +96,25 @@
           - "'GRANT SELECT, DELETE ON `data2`.*' in result.stdout"
       when: enable_check_mode == 'yes'
 
+    - name: Try to append invalid privileges
+      mysql_user:
+        <<: *mysql_params
+        name: '{{ user_name_4 }}'
+        password: '{{ user_password_4 }}'
+        priv: 'data1.*:INVALID/data2.*:SELECT'
+        append_privs: yes
+        state: present
+      check_mode: '{{ enable_check_mode }}'
+      register: result
+      ignore_errors: true
+
+    - name: Assert that there wasn't a change in privileges if check_mode is set to 'no'
+      assert:
+        that:
+          - "result.changed == false"
+          - "'Illegal privilege' in result.msg or 'You have an error in your SQL syntax;' in result.msg"
+      when: enable_check_mode == 'no'
+
     ##########
     # Clean up
     - name: Drop test databases
--- a/tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
+++ b/tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
@@ -111,7 +111,7 @@
     - name: Assert that there wasn't a change in privileges if check_mode is set to 'no'
       assert:
         that:
-          - "result.changed == false"
+          - result is failed
           - "'Error granting privileges' in result.msg"
       when: enable_check_mode == 'no'
 
--- a/changelogs/fragments/243-get-rid-of-privs-comparison.yml
+++ b/changelogs/fragments/243-get-rid-of-privs-comparison.yml
@@ -1,0 +1,2 @@
+breaking_changes:
+  - mysql_user - validate privileges using database engine directly (https://github.com/ansible-collections/community.mysql/issues/234 https://github.com/ansible-collections/community.mysql/pull/243). Do not validate privileges in this module anymore.
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -19,49 +19,6 @@ from ansible_collections.community.mysql.plugins.module_utils.mysql import (
 )
 
 
-EXTRA_PRIVS = ['ALL', 'ALL PRIVILEGES', 'GRANT', 'REQUIRESSL']
-
-# This list is kept for backwards compatibility after release 2.3.0,
-# see https://github.com/ansible-collections/community.mysql/issues/232 for details
-VALID_PRIVS = [
-    'CREATE', 'DROP', 'GRANT', 'GRANT OPTION',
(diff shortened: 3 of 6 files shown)


rsicart:
For the moment I removed all priv comparison code.


There's already some try/catch blocks which catch `mysql_client.Error` errors at `mysql_user` module level. I was thinking to raise a custom exception inside `privileges_grant()` but I'm not sure that we need more detail than this, catching the error at module level seems already enough. What do you think about that?

rsicart:
Also I'm new with `ansible-test`. I think that currently there's only mysql tests. How do you test mariadb with `ansible-test`? Do you have some tips for that?

I tried locally to set `mariadb_install: true` in setup_mysql target defaults, but I have some errors. It seems more complicated than that.

Andersson007:
@rsicart thanks for doing this!
I'm on vacation this week. @Jorge-Rodriguez , @bmalynovytch could you please take a look?

tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
```diff
@@ -96,6 +96,25 @@
...
+        append_privs: yes
+        state: present
+      check_mode: '{{ enable_check_mode }}'
+      register: result
+      ignore_errors: true
+
+    - name: Assert that there wasn't a change in privileges if check_mode is set to 'no'
+      assert:
+        that:
+          - "result.changed == false"
```
Andersson007:
```suggestion
          - result is failed
```
is this what we expect? Same for the other test task

Andersson007 (commented):
@rsicart thanks for working on this!

If we see CI green, it would be great

rsicart:
CI doesn't pass because of python2. I used a `raise ... from ...` which is python3 only syntax. I didn't fix it because perhaps it's time to go on and forget python 2?
This week I'm on holidays, I do not have my computer to fix this pr.

Andersson007:
> CI doesn't pass because of python2. I used a `raise ... from ...` which is python3 only syntax. I didn't fix it because perhaps it's time to go on and forget python 2? This week I'm on holidays, I do not have my computer to fix this pr.

Unfortunately we can't drop Python 2 support (at least in the near future) because this collection is included in Ansible package and it requires support of Python versions that supported by all supported Ansible versions:)
You could use just `module.fail_json()` instead, I think

Andersson007:
We also don't test against MariaDB. It would be really great to add it to our CI (i'd say it's vital to have it). I created an issue recently where help_wanted https://github.com/ansible-collections/community.mysql/issues/238. So at least for now manual testing against MariaDB will be required.

Andersson007:
@rsicart thanks for working on this!
When it's ready for review, please let us know and:
- Eliminate `Draft`
- Add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)
- Give feedback on my comment about CI:)

Thank you

rsicart:
I agree that having tests for mariadb is vital.

I just tested manually using docker on my computer and different version of mysql (5.6, 5.7, 8.0) and mariadb (10.4, 10.5).

I tried to add `INVALID` privilege to user `toto@%` using mysql client, in order to view error messages returned by the database.

Then I checked what kind of exception is raised by mysql modules (pymysql, mysqldb) in order to catch them, and I wrote integration tests to complete.

But doing that every time is annoying, automated testing for mariadb should be available in CI too!

tests/integration/targets/test_mysql_user/tasks/test_priv_append.yml
rsicart: Good catch

Jorge-Rodriguez:
> But doing that every time is annoying, automated testing for mariadb should be available in CI too!

There has been a plan to implement testing against MariaDB for quite some time now, but it is not quite a trivial task as it turns out so we haven't gotten around to it. Recent changes to the requirements for the testing versions have further delayed that task. I do have a mind to have a go at it if there's no progress on that front but I've been quite caught up otherwise. All in all, this is in the backlog.

Jorge-Rodriguez (approved):

Andersson007:
@rsicart thanks for working on this! Sorry for the later reply, was on PTO. Everything looks good. Thank you for adding the tests!
@Jorge-Rodriguez thanks for reviewing and merging!

Andersson007:
@rsicart @Jorge-Rodriguez @bmalynovytch i didn't notice that this PR marked with `breaking_change`. Actually this shouldn't affect users, right? What do you think if we release this in 3.0.0 (this or next week)?
I wouldn't wait for 4.0.0 as it hasn't been planned at all:)

Jorge-Rodriguez:
I'm good with 3.0.0

PR #266, opened 2021-12-22
Add default database "mysql" to mysql_user

### LINKED ISSUE #265: mysql_user: Not using the mysql database as default can break replication
##### SUMMARY
In replication setups that use filters on the mysql database (to have different privileges for single nodes or certain branches of the replication tree) using this module will break replication. Since the module is not using any database the statements will not be filtered out and you may try to change users that do not exists on the filtered replica.

I could open a PR to fix this, but i wanted to discuss how you want this solved as i see two possible solutions.

Root cause is that the mysql_user plugin does not set a database. This makes the user change statements bypass the mysql binlog/replication filters.

We could either
A) hardcode the default database for the mysql_user plugin to "mysql" as this is the place the permissions are stored anyway and usually the database that is filtered.
B) make the database setting configureable for the mysql_user plugin so the user can decide if this feature is needed.

I did test A on my setup and it works flawlessly but i could understand why one would want to make this configureable. Let me know which solution you prefer and i will open a PR.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
mysql_user

##### ANSIBLE VERSION
```
ansible [core 2.11.1] 
  config file = None
  configured module search path = ['/home/vagrant/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/vagrant/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/vagrant/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/vagrant/.local/bin/ansible
  python version = 3.8.10 (default, Nov 26 2021, 20:14:08) [GCC 9.3.0]
  jinja version = 2.10.1
  libyaml = True
```

##### COLLECTION VERSION
```
# /home/vagrant/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
containers.podman 1.7.1  
netbox.netbox     3.1.1  

# /home/vagrant/.local/lib/python3.8/site-packages/ansible_collections
Collection                    Version
----------------------------- -------
amazon.aws                    1.5.0  
ansible.netcommon             2.0.2  
ansible.posix                 1.2.0  
ansible.utils                 2.1.0  
ansible.windows               1.5.0  
arista.eos                    2.1.1  
awx.awx                       19.0.0 
azure.azcollection            1.5.0  
check_point.mgmt              2.0.0  
chocolatey.chocolatey         1.1.0  
cisco.aci                     2.0.0  
cisco.asa                     2.0.1  
cisco.intersight              1.0.15 
cisco.ios                     2.0.1  
cisco.iosxr                   2.1.0  
cisco.meraki                  2.2.1  
cisco.mso                     1.1.0  
cisco.nso                     1.0.3  
cisco.nxos                    2.2.0  
cisco.ucs                     1.6.0  
cloudscale_ch.cloud           2.1.0  
community.aws                 1.5.0  
community.azure               1.0.0  
community.crypto              1.6.2  
community.digitalocean        1.1.1  
community.docker              1.5.0  
community.fortios             1.0.0  
community.general             3.0.2  
community.google              1.0.0  
community.grafana             1.2.1  
community.hashi_vault         1.1.3  
community.hrobot              1.1.1  
community.kubernetes          1.2.1  
community.kubevirt            1.0.0  
community.libvirt             1.0.1  
community.mongodb             1.2.1  
community.mysql               2.1.0  
community.network             3.0.0  
community.okd                 1.1.2  
community.postgresql          1.2.0  
community.proxysql            1.0.0  
community.rabbitmq            1.0.3  
community.routeros            1.1.0  
community.skydive             1.0.0  
community.sops                1.0.6  
community.vmware              1.9.0  
community.windows             1.3.0  
community.zabbix              1.3.0  
containers.podman             1.5.0  
cyberark.conjur               1.1.0  
cyberark.pas                  1.0.6  
dellemc.enterprise_sonic      1.0.3  
dellemc.openmanage            3.3.0  
dellemc.os10                  1.1.1  
dellemc.os6                   1.0.7  
dellemc.os9                   1.0.4  
f5networks.f5_modules         1.9.0  
fortinet.fortimanager         2.0.2  
fortinet.fortios              2.0.1  
frr.frr                       1.0.3  
gluster.gluster               1.0.1  
google.cloud                  1.0.2  
hetzner.hcloud                1.4.3  
hpe.nimble                    1.1.3  
ibm.qradar                    1.0.3  
infinidat.infinibox           1.2.4  
inspur.sm                     1.1.4  
junipernetworks.junos         2.1.0  
kubernetes.core               1.2.1  
mellanox.onyx                 1.0.0  
netapp.aws                    21.2.0 
netapp.azure                  21.5.0 
netapp.cloudmanager           21.5.1 
netapp.elementsw              21.3.0 
netapp.ontap                  21.5.0 
netapp.um_info                21.5.0 
netapp_eseries.santricity     1.2.7  
netbox.netbox                 3.0.0  
ngine_io.cloudstack           2.1.0  
ngine_io.exoscale             1.0.0  
ngine_io.vultr                1.1.0  
openstack.cloud               1.4.0  
openvswitch.openvswitch       2.0.0  
ovirt.ovirt                   1.4.2  
purestorage.flasharray        1.8.0  
purestorage.flashblade        1.6.0  
sensu.sensu_go                1.9.4  
servicenow.servicenow         1.0.5  
splunk.es                     1.0.2  
t_systems_mms.icinga_director 1.16.0 
theforeman.foreman            2.0.1  
vyos.vyos                     2.2.0  
wti.remote                    1.0.1  
```

##### CONFIGURATION
no changed configuration

##### OS / ENVIRONMENT
Ubuntu 20.04.2 LTS @ ansible controller
RHEL7.9 with Percona Server 5.7.35 @ database hosts

##### STEPS TO REPRODUCE
You need a replication setup with at least three chained nodes.

writer -> intermediate -> backend

On the intermediate node you configure a replication filter ("binlog-ignore-db = mysql") to apply changes to the mysql database locally but not write them to the binlog for further replication (you could also do replicate-ignore-db to apply on this node).

The writer and intermediate pair share their mysql permissions. The backend has completely different permissions.

When you use the mysql_user plugin to change or remove a user the replication will break on the backend with it being unable to find the entries that need to be changed. 
When you use the mysql_user plugin to add a user it will add the user on the backend even though that is not supposed to happen.

##### EXPECTED RESULTS
Statements properly filtered by the intermediate node and not replicated to the backend.

##### ACTUAL RESULTS
Replication breaks or user is wrongfully added to the backend

d-rupp:
##### SUMMARY
Fixes #265 

This will change the database context of the executed statements to be the "mysql" database. Since permissions are stored in the "mysql" database anyway this should not change the behaviour of the module. But replication / binlog filters which rely on the current database will be able to filter the statements correctly afterwards. Prior to this change they were not executed in any database context and could not be filtered in any way by the existing methods in MySQL.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
mysql_user

##### ADDITIONAL INFORMATION
see related issue


Andersson007:
@d-rupp thanks for the PR! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)?

Andersson007:
@bmalynovytch @d-rupp @Jorge-Rodriguez @rsicart how about using the existing `login_db` argument instead of hardcoding?
It's not present in the module but it's used by the other modules (e.g. `mysql_info`) and has no default value. So i guess it should fit here.
What do you think?

bmalynovytch:
> @bmalynovytch @d-rupp @Jorge-Rodriguez @rsicart how about using the existing `login_db` argument instead of hardcoding? It's not present in the module but it's used by the other modules (e.g. `mysql_info`) and has no default value. So i guess it should fit here. What do you think?

I think it could confuse, as user management is global and shouldn't need to specify a database.
What @d-rupp explained is that there seem to be a bug (or at least different behaviors with or without database specified) with replication of user management statements.

plugins/modules/mysql_user.py
```diff
@@ -365,7 +365,7 @@ def main():
     ssl_key = module.params["client_key"]
     ssl_ca = module.params["ca_cert"]
     check_hostname = module.params["check_hostname"]
-    db = ''
+    db = 'mysql'
```
bmalynovytch: @d-rupp have you checked if `db` was really used in queries ?
Have you tested the module on your side to check if it really solves your issue ?

You can place a copy of the module in a `library` dir sitting next to your playbook to try.

Andersson007:
> > @bmalynovytch @d-rupp @Jorge-Rodriguez @rsicart how about using the existing `login_db` argument instead of hardcoding? It's not present in the module but it's used by the other modules (e.g. `mysql_info`) and has no default value. So i guess it should fit here. What do you think?
> 
> I think it could confuse, as user management is global and shouldn't need to specify a database. What @d-rupp explained is that there seem to be a bug (or at least different behaviors with or without database specified) with replication of user management statements.

@bmalynovytch as we already use `login_user`, `login_port`, `login_host`, and `login_password` with this module, would it be logical to also have `login_db`? I don't know maybe there can be situations when a certain `login_user` is allowed to connect to a certain DB? I think if it is a popular scenario, it would be already requested but anyway.
I'm not for or against, just trying to think logically:)

plugins/modules/mysql_user.py
d-rupp: Well, i did test the fix forward and backwards. 
I changed the file on my disk in the collection itself (```/home/vagrant/.local/lib/python3.8/site-packages/ansible_collections/community/mysql/plugins/modules/mysql_user.py``` in my case).

With the fix i applied a user creation, modification and deletion and replication on the filtered node kept running.

Without the fix it breaks with ```Last_SQL_Error: Error 'There is no such grant defined for user 'foo' on host '10.10.10.1'' on query. Default database: ''. Query: 'REVOKE PROCESS,REPLICATION CLIENT ON *.* FROM 'foo'@'10.10.10.1''```

bmalynovytch:
@Andersson007 I get your point, but `mysql_user` triggers global statements, there should be absolutely no need to specify a database. 
The case we're trying to fix is a bug either on MariaDB or MySQL, which handles differently global statements depending they were triggered in a "specific database context" or "no specific context". Both should work the same 🤔

plugins/modules/mysql_user.py
d-rupp: Small error: i apply a modification to an existing user, a creation of a new user, a modification to the new user and deletion of the new user in that order.
The first modification of the existing user already breaks replication as the users permissions are not present on the filtered replica.
Now if i run the test again (without the fix) without modifying the existing user, but creating a new user -> the change will get applied to the replica even though it should NOT be applied and filtered out.

d-rupp:
Thats why i asked what solution you prefer, you could make a case for either one 😅

My experience as a 15 years DBA tells me to just stick it to the mysql database. My other side tells me to make all the the things configureable. And then there is the first one again that says you should not give any options that do not matter anyway and just confuse people 😂

changelog fragments added (i hope i did it correctly)

Andersson007:
> @Andersson007 I get your point, but `mysql_user` triggers global statements, there should be absolutely no need to specify a database. The case we're trying to fix is a bug either on MariaDB or MySQL, which handles differently global statements depending they were triggered in a "specific database context" or "no specific context". Both should work the same thinking

I'm OK with the hardcoding if it's tested and safe:)

changelogs/fragments/266-default-database-for-mysql-user
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+- mysql_user - set default database context for the queries to be the "mysql" database. This way replication/binlog filters can catch the statements.
```
Andersson007:
```suggestion
- mysql_user - set default database context for the queries to be the ``mysql`` database. This way replication/binlog filters can catch the statements (https://github.com/ansible-collections/community.mysql/issues/265).
```

Andersson007:
@d-rupp @bmalynovytch can this break anything? I.e. some specific permission configurations?

Andersson007:
> @d-rupp @bmalynovytch can this break anything? I.e. some specific permission configurations?

Can connections to `mysql` technically be forbidden? If yes, `login_db` feels like the only solution

d-rupp:
> > @d-rupp @bmalynovytch can this break anything? I.e. some specific permission configurations?
> 
> Can connections to `mysql` technically be forbidden? If yes, `login_db` feels like the only solution

This should not break anything. 

And i just tried some of this. You could technically create a user that just has the permissions to create a user. But you could not assign any grants to that user then, so imho this is pretty pointless. And once you give the user the option to give grants to a restricted set of databases he can not create users anymore. And if you give him the global grant permissions he can read all databases. Pretty wild stuff, i don´t really see a reason why you should do these.

d-rupp:
I guess if you want to be on the "very very safe" side we make this configureable.

bmalynovytch:
"Configurable" could be a bool: "force connection to be established on db mysql (and not just no database selected)"

d-rupp:
Reverted prior change, added a configureable boolean "db_context",a dded documentation, updated changelog.

plugins/modules/mysql_user.py
```diff
@@ -74,6 +74,12 @@
       - Whether binary logging should be enabled or disabled for the connection.
     type: bool
     default: yes
+  db_context:
```
bmalynovytch: not completely convinced by the name, as it doesn't seem clear at first glance, but can't find something better right now 🤔

d-rupp: Yeah, me neither... as usual, one of the 2 hardest things in IT... i had login_db in my first draft but i wasn´t convinced. I´m up for suggestions.

plugins/modules/mysql_user.py
```diff
@@ -74,6 +74,12 @@
...
+    description:
+      - Sets the mysql system database as context for the executed statements. Useful if you use
```
```suggestion
      - Sets the С(mysql) system database as context for the executed statements (it will be used as a database to connect to). Useful if you use 
```
Or something like that as `mysql` will be passed to the driver's connection function.

I would also provide more context, i.e. that the users can get the error described in the issue.

plugins/modules/mysql_user.py
```diff
@@ -366,6 +373,7 @@ def main():
     ssl_ca = module.params["ca_cert"]
     check_hostname = module.params["check_hostname"]
     db = ''
+    db_context = module.params["db_context"]
```
```suggestion
    if module.params["db_context"]:
        db = 'mysql'
```
to keep it closer, would make things more readable

plugins/modules/mysql_user.py
```diff
@@ -74,6 +74,12 @@
       - Whether binary logging should be enabled or disabled for the connection.
     type: bool
     default: yes
+  db_context:
+    description:
+      - Sets the mysql system database as context for the executed statements. Useful if you use 
+        binlog / replication filters in MySQL.
+    type: bool
+    default: no
```
Andersson007, line 88:
```suggestion
    default: no
    version_added: '3.1.0'
```

plugins/modules/mysql_user.py
```diff
@@ -379,6 +387,9 @@ def main():
 
     if mysql_driver is None:
         module.fail_json(msg=mysql_driver_fail_msg)
+    
+    if db_context:
+        db = 'mysql'
```
```suggestion
```
See the suggestion above

plugins/modules/mysql_user.py
bmalynovytch: `force_context` ?

Andersson007:
could you also please add the new parameter with `yes` value  (after you choose a final name) to any related task within `tests/integration/targets/mysql_replication/tasks` ?

plugins/modules/mysql_user.py
Andersson007: `db_context` works for me, though I'd be happy with any name you'll agree on:)

d-rupp:
> could you also please add the new parameter with `yes` value (after you choose a final name) to any related task within `tests/integration/targets/mysql_replication/tasks` ?

Not sure if i did this correctly, please check

plugins/modules/mysql_user.py
```diff
@@ -74,6 +74,19 @@
...
+  force_context:
+    description:
+      - Sets the С(mysql) system database as context for the executed statements (it will be used
+        as a database to connect to). Useful if you use binlog / replication filters in MySQL as
+        per default the statements can not be caught by a binlog / replication filter, they require
+        a database to be set to work.
+      - see U(https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#option_mysqld_binlog-ignore-db)
+        for a description on how binlog filters work (filtering on the primary).
+      - see U(https://dev.mysql.com/doc/refman/8.0/en/replication-options-replica.html#option_mysqld_replicate-ignore-db)
+        for a description on how replication filters work (filtering on the replica).
```
```suggestion
      - See U(https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#option_mysqld_binlog-ignore-db)
        for a description on how binlog filters work (filtering on the primary).
      - See U(https://dev.mysql.com/doc/refman/8.0/en/replication-options-replica.html#option_mysqld_replicate-ignore-db)
        for a description on how replication filters work (filtering on the replica).
```

plugins/modules/mysql_user.py
```diff
@@ -74,6 +74,19 @@
       - Whether binary logging should be enabled or disabled for the connection.
     type: bool
     default: yes
+  force_context:
+    description:
+      - Sets the С(mysql) system database as context for the executed statements (it will be used
+        as a database to connect to). Useful if you use binlog / replication filters in MySQL as
+        per default the statements can not be caught by a binlog / replication filter, they require
+        a database to be set to work.
```
```suggestion
        a database to be set to work, otherwise the replication can break down.
```
Or something like that (not sure if we also need a period after `otherwise`)

Andersson007 (commented):
Oh, sorry, could you also please add `force_option: yes` to somewhere within `tests/integration/targets/mysql_user/tasks` as the same as you did for `mysql_replication`. Things got mixed in my head:)

Andersson007 (approved):
LGTM

Andersson007:
@bmalynovytch are you OK with the PR as it is now?

bmalynovytch:
I'm just wondering why code coverage is complaining 🤔 
Shouldn't we also add specific tests with the new option with either values ? (at least basic user management, don't need to bother with replication I guess)

Andersson007:
> I'm just wondering why code coverage is complaining thinking Shouldn't we also add specific tests with the new option with either values ? (at least basic user management, don't need to bother with replication I guess)

Do you mean with `no` also? Not sure as it is default

Andersson007:
@d-rupp could you please set one of the added tests with `no` explicitly?

bmalynovytch:
In fact, we need to ensure that running any test with either `yes` or `no` produces the same result (except replicating user management statement, which I guess isn't covered right now)

d-rupp:
I will pick this up again next week or next year as this is all the time i had today. Will try to write a proper test case, i have an idea. And i don´t think we are in a rush.

Andersson007:
@d-rupp hi, happy new year! Thank you for working on this!
You can run tests locally before pushing to save your time:

`ansible-test integration test_mysql_user --docker` I usually also add `-vvv > ~/test.log`
`ansible-test sanity /path/to/file --docker`

You need to have docker and ansible-core installed / clonned ansible/ansible repo in your system, see https://github.com/ansible/community-docs/blob/main/create_pr_quick_start_guide.rst for details.

If any question, feel free to ask

d-rupp:
Thanks, that was probably a lot quicker 😅 Tests work locally, hope that they will work on Github too now.

d-rupp:
Ok, so MariaDB does things a bit differently, should be integrated now. Hopefully this is it.

changelogs/fragments/266-default-database-for-mysql-user.yml
```diff
@@ -0,0 +1,2 @@
+minor_change:
```
Andersson007:
```suggestion
minor_changes:
```
this should fix sanity tests

d-rupp: *facepalm* thanks, i was wondering whats wrong...

Andersson007: you're welcome!

changelogs/fragments/266-default-database-for-mysql-user.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- mysql_user - added the force_context boolean option to set the default database context for the queries to be the ``mysql`` database. This way replication/binlog filters can catch the statements (https://github.com/ansible-collections/community.mysql/issues/265).
```
Andersson007:
```suggestion
- mysql_user - added the ``force_context`` boolean option to set the default database context for the queries to be the ``mysql`` database. This way replication/binlog filters can catch the statements (https://github.com/ansible-collections/community.mysql/issues/265).
```
to highlight it a bit in the changelog

tests/integration/targets/test_mysql_user/tasks/issue-265.yml
```diff
@@ -0,0 +1,168 @@
...
+        name: '{{user_name_1}}'
+        password: '{{user_password_1}}'
+        state: absent
+        force_context: no
+      register: result
+
+    - name: assert output message mysql user was removed
+      assert:
+        that:
+          - "result.changed == true"
```
Andersson007, line 129:
```suggestion
          - "result.changed == true"
```
No need to change, just another form of the same can be just
```
- result is changed
```
it's a question of taste, i think

d-rupp: I prefer the latter in my playbooks but i actually just copied most of that from the other playbooks 😅

Andersson007: :)

Andersson007 (approved):
LGTM

Andersson007:
cc @bmalynovytch @rsicart @Jorge-Rodriguez 
there will be one sanity failing - it's unrelated, feel free to review

Andersson007:
closed reopened to restart CI

Andersson007:
@d-rupp everything is green now, thank you!
Let's wait until Monday for someone else's approval. If there are at least no objections, I'll merge the PR on Monday.

Andersson007:
closed reopened it because of the hanging codecov test
@bmalynovytch @Jorge-Rodriguez @rsicart there's the last chance to take a look at the PR today:) I'm gonna merge it at 15 UTC if there are no objections. If no time, feel free to ignore:)

rsicart:
Is that normal that sanity check for ansible 2.9 does not pass?

Otherwise code seems ok for me.

Andersson007:
> Is that normal that sanity check for ansible 2.9 does not pass?
> 
> Otherwise code seems ok for me.

Everything seems to be green to me. Except codecov but it's not an obstacle in this case, I think.

rsicart:
> > Is that normal that sanity check for ansible 2.9 does not pass?
> > Otherwise code seems ok for me.
> 
> Everything seems to be green to me. Except codecov but it's not an obstacle in this case, I think.

Ah yes you're right, I saw an old sanity check in "Checks" tab. And stable-2.9 is currently ok.

rsicart (approved):
LGTM

Andersson007:
@d-rupp thanks for the patch! If you'd like to contribute more, we have a lot of issues / PRs to solve / review:)
@bmalynovytch @rsicart thanks for reviewing!

PR #286, opened 2022-02-12
Add IF EXISTS to DROP USER / ROLE command

TodoInTX:
##### SUMMARY
Add IF EXISTS clause to DROP USER statement

There exists the possibility for a race condition that breaks certain (circular) replication configurations when DROP USER is executed on multiple nodes in the replica set.  Adding IF EXISTS avoids the need to use `sql_log_bin : no` making the statement always replication safe.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
module_utils/user.py

### DIFF
--- a/changelogs/fragments/286-mysql_drop_user_role_if_exists.yml
+++ b/changelogs/fragments/286-mysql_drop_user_role_if_exists.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - mysql_role and mysql_user Adds IF EXISTS clause to DROP ROLE and DROP USER statements to make them replication safe on MySQL 5.7+ and MariaDB 10.1+
--- a/changelogs/fragments/286-mysql_drop_user_role_if_exists.yml
+++ b/changelogs/fragments/286-mysql_drop_user_role_if_exists.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - mysql_role and mysql_user Adds IF EXISTS clause to DROP ROLE and DROP USER statements to make them replication safe on MySQL 5.7+ and MariaDB 10.1+
+  - mysql_role and mysql_user - adds IF EXISTS clause to DROP ROLE and DROP USER statements to make them replication safe on MySQL 5.7+ and MariaDB 10.1+ (https://github.com/ansible-collections/community.mysql/pull/286).
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -368,7 +368,7 @@ def user_delete(cursor, user, host, host_all, check_mode):
         hostnames = [host]
 
     for hostname in hostnames:
-        cursor.execute("DROP USER /*!50700 IF EXISTS */ /*M!100100 IF EXISTS */  %s@%s", (user, hostname))
+        cursor.execute("DROP USER /*!50708 IF EXISTS */ /*M!100103 IF EXISTS */  %s@%s", (user, hostname))
 
     return True
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -368,7 +368,7 @@ def user_delete(cursor, user, host, host_all, check_mode):
         hostnames = [host]
 
     for hostname in hostnames:
-        cursor.execute("DROP USER %s@%s", (user, hostname))
+        cursor.execute("DROP USER IF EXIST %s@%s", (user, hostname))
 
     return True
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -368,7 +368,7 @@ def user_delete(cursor, user, host, host_all, check_mode):
         hostnames = [host]
 
     for hostname in hostnames:
-            cursor.execute("DROP USER /*!50700 IF EXISTS */ /*M!100103 IF EXISTS */  %s@%s", (user, hostname))
+            cursor.execute("DROP USER /*!50700 IF EXISTS */ /*M!100100 IF EXISTS */  %s@%s", (user, hostname))
 
     return True
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -368,7 +368,7 @@ def user_delete(cursor, user, host, host_all, check_mode):
         hostnames = [host]
 
     for hostname in hostnames:
-        cursor.execute("DROP USER IF EXISTS %s@%s", (user, hostname))
+            cursor.execute("DROP USER /*!50700 IF EXISTS */ /*M!100103 IF EXISTS */  %s@%s", (user, hostname))
 
     return True
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -368,7 +368,7 @@ def user_delete(cursor, user, host, host_all, check_mode):
         hostnames = [host]
 
     for hostname in hostnames:
-        cursor.execute("DROP USER IF EXIST %s@%s", (user, hostname))
+        cursor.execute("DROP USER IF EXISTS %s@%s", (user, hostname))
 
     return True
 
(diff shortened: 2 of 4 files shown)


rsicart:
Hi @TodoInTX 
Thanks for contributing :)

Did you check if there are some other 'DROP USER' statements in the code?

Code seems good, but I'd like to see all tests passing before.

TodoInTX:
Hi R. Sicart

Yes, this appears to be the only "DROP USER" instance in the code.

~/GitHub/community.mysql$ grep -R "DROP USER" .
./plugins/module_utils/user.py:        cursor.execute("DROP USER IF EXISTS
%s@%s", (user, hostname))

I do see DROP ROLE and DROP DATABASE statements that might also suffer from
the same race condition but I have not encountered those.

Andersson007:
@TodoInTX hi, welcome to the project and thank you for the fix!
Could you also please add a changelog fragment https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment ?

@rsicart , after @TodoInTX adds the fragment, feel free to merge the patch if you consider it useful.

Andersson007:
ps. we should be sure this won't break backwards compatibility and will work in mysql and mariadb of different supported versions (i.e. the case is covered by integration tests)

TodoInTX:
Anderson,

MySQL supports the IF EXISTS clauses on DROP USER statements in 5.7.8+ and MariaDB does so for DROP ROLE and USER in 10.1.3+.  In my previous commit I've enclosed the IF EXISTS clauses in version specific comments such that MySQL and MariaDB versions without support for this clause will ignore it. 

Currently the comments include all versions of 5.7 and 10.1. That would be a problem for someone running pre-GA minor version of already very old major versions. I can narrow them further if that would be preferred. 

See: 
https://mariadb.com/kb/en/comment-syntax/
https://mariadb.com/kb/en/mariadb-1013-release-notes/ 

https://dev.mysql.com/doc/refman/5.7/en/comments.html
https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-8.html#:~:text=Bugs%20Fixed-,Account%20Management%20Notes,-The%20CREATE%20USER

Changelog fragment has been added.  Sorry for the omission.

changelogs/fragments/286-mysql_drop_user_role_if_exists.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - mysql_role and mysql_user Adds IF EXISTS clause to DROP ROLE and DROP USER statements to make them replication safe on MySQL 5.7+ and MariaDB 10.1+
```
Andersson007:
```suggestion
  - mysql_role and mysql_user - adds IF EXISTS clause to DROP ROLE and DROP USER statements to make them replication safe on MySQL 5.7+ and MariaDB 10.1+ (https://github.com/ansible-collections/community.mysql/pull/286).
```

Andersson007 (commented):
@TodoInTX thanks for explaining! I'm not a user, so if @rsicart is OK with the patch safety, it would be enough. Could you please fix the fragment a bit?

rsicart:
@TodoInTX I think it'd be better to define specific minimal versions, because you know already them and to avoid surprises.

I'm not familiar with those type of sql commands in comments, but I checked official docs you linked, and tested locally with mysql 5.7.x and mariadb 10.1.x docker containers using `DROP` statements in this PR and commands work fine.

That being said, @Andersson007 I'm not sure about breaking backwards compatibility. Is there someone in the room with more experience using mysql/mariadb? Currently integration tests for mysql_user target pass, but for [mysql_role target are not enabled for mysql 5.7 or mariadb](https://github.com/ansible-collections/community.mysql/blob/main/tests/integration/targets/test_mysql_role/tasks/mysql_role_initial.yml#L33).

TodoInTX:
Since the DROP ROLE command does not exist in any version of MySQL, only MariaDB 10.0.5+, I've constructed the DROP ROLE command such that no part of the DROP ROLE command is executed on MySQL. The DROP ROLE command will execute on MariaDB 10.0.5 - 10.1.2 without the IF EXISTS clause and will use IF EXISTS clause on 10.1.3 or later.

rsicart:
> Since the DROP ROLE command does not exist in any version of MySQL

Please can you confirm that? I can see DROP ROLE command and other related role commands in [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/roles.html)

Andersson007:
Can we use just try except? In other words:
```
try:
    DROP ROLE role IF EXISTS ...
except
    DROP ROLE role ...
```
?

Andersson007:
~~the suggested change will probable break things~~
```
        cursor.execute("DROP USER /*!50708 IF EXISTS */ /*M!100103 IF EXISTS */  %s@%s", (user, hostname))

    return True
```
~~because it will report True in all cases~~

Andersson007:
~~We have the following code in `mysql_user`:~~
```
    elif state == "absent":
        if user_exists(cursor, user, host, host_all):
            changed = user_delete(cursor, user, host, host_all, module.check_mode)
            msg = "User deleted"
        else:
            changed = False
            msg = "User doesn't exist"
```
~~If we use `user_exists` in `user_delete` as a condition before running DROP command, will it help? I.e.~~
```
if user_exist(...):
   self.cursor.execute('DROP ROLE %s', (self.name,))
   return True
else:
   return False
```
~~What do you folks think?~~
UPDATED: i think it won't help

Andersson007:
~~I think we could also check what the driver returns and determine what to return (True/False) based on the message (IIRC it's cursor.statusmessage or maybe something else~~

plugins/module_utils/user.py
```diff
@@ -368,7 +368,7 @@ def user_delete(cursor, user, host, host_all, check_mode):
         hostnames = [host]
 
     for hostname in hostnames:
-        cursor.execute("DROP USER %s@%s", (user, hostname))
+        cursor.execute("DROP USER /*!50708 IF EXISTS */ /*M!100103 IF EXISTS */  %s@%s", (user, hostname))
```
Andersson007, line 371:
```suggestion
        try:
            cursor.execute("DROP USER IF EXISTS  %s@%s", (user, hostname))
        except:
            cursor.execute("DROP USER %s@%s", (user, hostname))
```

Andersson007: @TodoInTX @rsicart sorry for the noise, i didn't sleep well, so please ignore all my comments above. This is my proposal.

Andersson007: same for role

TodoInTX:
> > Since the DROP ROLE command does not exist in any version of MySQL
> 
> Please can you confirm that? I can see DROP ROLE command and other related role commands in [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/roles.html)

Sorry, I didn't find it in the 5.7 manual I was focused on and didn't check there.  My mistake.

plugins/module_utils/user.py
rsicart: no problem @Andersson007 :)
and thanks for helping with the review!

Personally I'm more comfortable with this solution too.

@TodoInTX what do you think about that proposal?

rsicart:
> Sorry, I didn't find it in the 5.7 manual I was focused on and didn't check there. My mistake.

No problem, we are here for that too!

plugins/module_utils/user.py
rsicart: @TodoInTX what do you think about the changes proposed earlier in this conversation?

Andersson007:
@TodoInTX please take a look at https://github.com/ansible-collections/community.mysql/pull/286#discussion_r811263631 or please let us know if someone else can complete the changes

Andersson007:
https://github.com/ansible-collections/community.mysql/pull/307

Andersson007:
@TodoInTX thanks for reporting the issue!

PR #290, opened 2022-02-28
Enable and fix integration tests for devel

### LINKED ISSUE #229: Fix and turn on integration tests against devel
##### SUMMARY
After this change in the default container https://github.com/ansible-collections/community.mysql/pull/226#discussion_r725619563 CI against `devel` has started failing.

They were turned off in https://github.com/ansible-collections/community.mysql/pull/228 (it's not mandatory but we have to fix it until branch `stable-2.13` will be created.

Fix and turn it on in a separate PR

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
.github/workflows/ansible-test-plugins.yml

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.mysql/issues/229

Enable and fix integration tests for devel.

The Ansible default container some time ago updated ubuntu 18.04 ->  ubuntu 20.04. Unfortunately mariadb does not provide package 10.2 for this distribution, so i changed it to 10.3
- https://downloads.mariadb.com/MariaDB/mariadb-10.2/repo/ubuntu/
- https://downloads.mariadb.com/MariaDB/mariadb-10.3/repo/ubuntu/

### DIFF
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -120,7 +120,7 @@ jobs:
         run: "sed -i 's/^python_packages:.*/python_packages: [${{ matrix.connector }}]/' ${{ env.connector_version_file }}"
 
       - name: Run integration tests
-        run: ansible-test integration --docker -v --color --retry-on-error --continue-on-error --python ${{ matrix.python }} --diff --coverage
+        run: ansible-test integration --docker ubuntu1804 -v --color --retry-on-error --continue-on-error --python ${{ matrix.python }} --diff --coverage
         working-directory: ./ansible_collections/community/mysql
 
       - name: Generate coverage report.
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -65,7 +65,7 @@ jobs:
           - stable-2.10
           - stable-2.11
           - stable-2.12
-          #- devel
+          - devel
         python:
           - 3.6
           - 3.8
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -120,7 +120,7 @@ jobs:
         run: "sed -i 's/^python_packages:.*/python_packages: [${{ matrix.connector }}]/' ${{ env.connector_version_file }}"
 
       - name: Run integration tests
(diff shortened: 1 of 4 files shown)


Andersson007:
ready for review

tests/integration/targets/setup_mysql/vars/main.yml
```diff
@@ -15,6 +15,7 @@ python_packages: [pymysql == 0.9.3]
 install_prereqs:
   - libaio1
   - libnuma1
+  - libncurses5
```
rsicart, line 18: Why is that package needed?

.github/workflows/ansible-test-plugins.yml
```diff
@@ -58,14 +58,14 @@ jobs:
...
-          - mariadb_10.2.37
+          - mariadb_10.3.34
           - mariadb_10.5.9
         ansible:
           - stable-2.9
           - stable-2.10
           - stable-2.11
           - stable-2.12
-          #- devel
+          - devel
```
rsicart, line 68: What's the purpose of devel branch? I ask because since I'm in the project we never used a devel branch, only main.

rsicart:
Also I think that for this PR a fragment is not needed. So you confirm?

.github/workflows/ansible-test-plugins.yml
Andersson007: this is the current devel branch of ansible-core. Shortly `stable-2.13` will be branched from it and if our code doesn't work with `devel` now, it will not work with the `stable-2.13` as well, i.e. it won't work with ansible-core 2.13 and Ansible 6 respectively. So this is why i consider important to test against it at least some time before the next ansible release (May, IIRC)

tests/integration/targets/setup_mysql/vars/main.yml
Andersson007: the default test container was upgraded to ubuntu2004 as well as some libraries were removed, so collection maintainers should install them manually if they need them. If we remove this line, `dbdeployer` will fail:) I've just checked it once more.

Andersson007:
> Also I think that for this PR a fragment is not needed. So you confirm?

Yep, changelog fragments are not needed for CI, doc, new modules / plugins PRs. Thanks for asking!

rsicart:
LGTM

Andersson007:
@rsicart if you're OK with this PR, feel free to merge it

rsicart:
Done!

Andersson007:
@rsicart thank you much for reviewing and mergning!

PR #316, opened 2022-03-22
Fix CI for roles

Andersson007:
##### SUMMARY
Fix CI for roles.
They've recently started failing. As we don't have roles anyway, if the first attempt doesn't help, I'll just remove the file.

### DIFF
--- a/.github/workflows/ansible-test-roles.yml
+++ b/.github/workflows/ansible-test-roles.yml
@@ -24,13 +24,24 @@ jobs:
           - 2.0.12
         ansible:
           - stable-2.9
-          ### it looks like there's errors for 2.10+ with ansible-lint (https://github.com/ansible/ansible-lint/pull/878)
-          ### and molecule (_maybe_ relating to https://github.com/ansible-community/molecule/pull/2547)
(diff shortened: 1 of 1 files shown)


rsicart:
😳
What's the purpose of this pipeline? Testing Ansible installation with different available python versions?

Andersson007:
> flushed What's the purpose of this pipeline? Testing Ansible installation with different available python versions?

Thanks for the question! As collections can also contain roles, i think it was originally added just "in advance". There's a commented command at the bottom of the file. We don't have any roles at the moment but, I guess, if once we have some, we'll need to uncomment the command.
I was thinking about removing the file or renaming it but it doesn't require a lot of resources and i decided to leave it there. Might be useful in the future.
What do you think? I'm not against of removing but, on the other hand, it doesn't require a lot to have it there.

rsicart:
Ahhh yes, thanks. We are talking about ansible roles, it makes sense :) . I agree, let's keep it.
LGTM 👍

rsicart (approved):

Andersson007:
@rsicart thanks for reviewing and merging!

PR #322, opened 2022-03-25
mysql_query: fix false change reports when IF NOT EXISTS clause is used

### LINKED ISSUE #268: mysql_query returns changed: true when nothing is executed.
##### SUMMARY
I used the mysql_query with some "CREATE TABLE IF NOT EXISTS" statements.  
The second time I run the query, it reports "changed" although nothing is executed since the tables are already created. The `mysql_query` module has the "changed: true" hardcoded in the return values simply by looking if some keywords are present. However, as you can see it's a bit more complex than that. 

Let's see this SQL query:
```sql
CREATE TABLE IF NOT EXISTS `instructions` (
  `id` int(11) NOT NULL,
  `instructions` text,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
If we execute this statement in the mysql/mariadb CLI client, we will see this:

```
CREATE TABLE IF NOT EXISTS `instructions` (`id` int(11) NOT NULL, `instructions` text, PRIMARY KEY (`id`) ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
Query OK, 0 rows affected, 1 warning (0.000 sec)
```
The warning of course is that the table already exists.

```
Note (Code 1050): Table 'instructions' already exists
```
When I execute the statement for a table that doesn't exist, we see this: 

```sql
CREATE TABLE IF NOT EXISTS `instructions3` (`id` int(11) NOT NULL, `instructions` text, PRIMARY KEY (`id`) ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
```
Query OK, 0 rows affected (0.012 sec)
```
We have no warnings, the query IS OK, so, we can safely _assume_ that the table is created. 
I don't know if you have discussed it in the past, but in my opinion the code should be a bit more selective on when to return "Changed" and when not. I classify this as a bug, but obviously if you feel that it's a feature request, I am more than fine with it.  

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
mysql_query

##### ANSIBLE VERSION

```paste below
ansible 2.9.27
```

##### COLLECTION VERSION

```paste below
ansible-collection-community-mysql-2.1.0-1.fc34.noarch
```

##### STEPS TO REPRODUCE

```yaml
- name: "Create the tables of {{ db_name }}"
  mysql_query:
    ca_cert: "{{ db_ca_cert | default(omit) }}"
    client_cert: "{{ db_client_cert | default(omit) }}"
    client_key: "{{ db_client_key | default(omit) }}"
    login_host: "{{ db_host }}"
    login_port: "{{ db_host_port if db_host != 'localhost' else omit }}"
    login_unix_socket: "{{ db_login_socket | default(omit) }}"
    login_user: "{{ db_admin_user }}"
    login_password: "{{ db_admin_pass }}"
    login_db: "{{ db_name }}"
    #query: "{{ lookup('template', 'schema.sql.j2') | replace('\n', '') | trim }}"
    query: "CREATE TABLE IF NOT EXISTS `instructions2` (`id` int(11) NOT NULL, `instructions` text, PRIMARY KEY (`id`)) ENGINE=InnoDB DEFAULT CHARSET=utf8;"
  register: sql_result
```

##### EXPECTED RESULTS
Module should return OK when no action is taken from the DB server when we execute DDL statements. 

##### ACTUAL RESULTS

```paste below
TASK [demo : Create the tables of demo] ***********************************************************************************************************************************
changed: [server.example.com]

TASK [demo : debug] **********************************************************************************************************************************************************
ok: [server.example.com] => 
  sql_result:
    changed: true
    executed_queries:
    - CREATE TABLE IF NOT EXISTS `instructions2` (  `id` int(11) NOT NULL,  `instructions` text,  PRIMARY KEY (`id`)) ENGINE=InnoDB DEFAULT CHARSET=utf8;
    failed: false
    query_result:
    - []
    rowcount:
    - 0

```

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.mysql/issues/268

When there's a statement like `create table IF EXISTS ...` it always reports `changed: yes`.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/changelogs/fragments/322-mysql_query_fix_false_change_report.yml
+++ b/changelogs/fragments/322-mysql_query_fix_false_change_report.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- mysql_query - fix false change reports when ``IF EXISTS/IF NOT EXISTS`` clause is used (https://github.com/ansible-collections/community.mysql/issues/268).
--- a/plugins/modules/mysql_query.py
+++ b/plugins/modules/mysql_query.py
@@ -22,9 +22,10 @@ options:
     description:
     - SQL query to run. Multiple queries can be passed using YAML list syntax.
     - Must be a string or YAML list containing strings.
-    - Note that if you use the C(IF EXISTS/IF NOT EXISTS) clauses in your query,
-      the module will report that the state has been changed even if it has not
-      when C(mysqlclient) connector is used.
+    - Note that if you use the C(IF EXISTS/IF NOT EXISTS) clauses in your query
+      and C(mysqlclient) connector, the module will report that
+      the state has been changed even if it has not. If it is important in your
+      workflow, use the C(PyMySQL) connector instead.
     type: raw
     required: yes
   positional_args:
--- a/plugins/modules/mysql_query.py
+++ b/plugins/modules/mysql_query.py
@@ -103,6 +103,8 @@ rowcount:
     sample: [5, 1]
 '''
 
+import warnings
+
 from ansible.module_utils.basic import AnsibleModule
 from ansible_collections.community.mysql.plugins.module_utils.mysql import (
     mysql_connect,
@@ -196,9 +198,21 @@ def main():
     executed_queries = []
     rowcount = []
 
+    already_exists = False
     for q in query:
         try:
-            cursor.execute(q, arguments)
+            with warnings.catch_warnings():
+                warnings.filterwarnings(action='error',
+                                        message='.*already exists*',
+                                        category=mysql_driver.Warning)
+
+                try:
+                    cursor.execute(q, arguments)
+                except mysql_driver.Warning:
+                    # When something is run with IF NOT EXISTS
+                    # and there's "already exists" MySQL warning,
+                    # set the flag as True
+                    already_exists = True
 
         except Exception as e:
             if not autocommit:
@@ -208,7 +222,8 @@ def main():
             module.fail_json(msg="Cannot execute SQL '%s' args [%s]: %s" % (q, arguments, to_native(e)))
 
         try:
-            query_result.append([dict(row) for row in cursor.fetchall()])
+            if not already_exists:
+                query_result.append([dict(row) for row in cursor.fetchall()])
 
         except Exception as e:
             if not autocommit:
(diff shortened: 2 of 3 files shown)


Andersson007:
@rsicart @itcultus @Jorge-Rodriguez @bmalynovytch  i don't know maybe there's a more sophisticated and safe way of fixing this.. any ideas?

rsicart:
I didn't know the `warnings` module.

Personally (because I'm not familiar with this module) I would try to solve that using `try/except`, catching `mysql_driver.Warning` and looking into exception message for the string `already exist`. In case the string does not match, re-raise the exception if needed.

But I'm not sure that this would work or if it would be more elegant that your proposed solution using `warnings` module.

Andersson007:
@rsicart thanks, I'll try the way you mentioned!
Psycopg2 (for postgres) has the `statusmessage` attribute which would be also cool to have in mysql drivers, but it seems to be not the case.. I'll try to work with `Warning` directly and let you know

Andersson007:
there's a problem with mysqlclient again :(

Andersson007:
Heh, i'm afraid this won't work with `mysqldb/mysqlclient` because it doesn't raise the warning in this case... This is a great feature of pymysql, see https://github.com/PyMySQL/PyMySQL/issues/688 (which is considered as a bug in some cases).
Also i think we should use the `warnings` module as we are interesting in catching a particular warning.

@rsicart what do you think if we should implement this for `PyMySQL` **only** or we should close the PR and change the doc respectively?

rsicart:
> @rsicart what do you think if we should implement this for `PyMySQL` **only** or we should close the PR and change the doc respectively?

I think we should do both 🙂

Andersson007:
ready_for_review

Andersson007:
@rsicart i'd be grateful for code review here

rsicart:
LGTM

rsicart (approved):

Andersson007:
@rsicart thank you for reviewing!

rsicart:
np :)

I totally forgot about this PR, thanks for pinging me and for the fix ^^

PR #333, opened 2022-04-12
mysql_user, mysql_role: add argument subtract_privs to revoke privileges explicitly

### LINKED ISSUE #331: mysql_user, mysql_role: add option to revoke privileges explicitly
#### SUMMARY
When updating a role, its privileges are either added or replaced, depending on the `append_privs` argument. I would like to add an argument to revoke privileges explicitly.

##### USECASE
I read the desired privileges from a dynamic source. To ensure no privileges are revoked in case the dynamic source accidentally returns incomplete information, I need to revoke privileges explicitly.

#### ISSUE TYPE
- Feature Idea

#### COMPONENT NAME
modules mysql_role, mysql_user

#### IMPLEMENTATION IDEAS
I propose 3 alternative ways to implement that:

##### 1) add argument subtract_privs
Add a new boolean argument `subtract_privs` (default `no`, mutual conflict with `append_privs`). If enabled, the module would revoke the privileges specified by `priv` (without granting any privileges).

##### 2) add argument revoke_privs
Add a new argument `revoke_privs` which takes the same format as `privs`. If set, all privileges in `revoke_privs` that are not in `privs` will be revoked.
`privs` and `append_privs` continue to work as before.

##### 3) add argument privilege_state, deprecate option append_privs
Add a new argument `privilege_state` with three possible values:
- `exact` (default, like previous behavior with `append_privs: no`)
- `granted` (like previous behaviour with `append_privs: yes`)
- `revoked` (revoke all privileges specified by `priv`)

This options should mutually conflict with `append_privs`. `privilege_state: granted` should be used instead of `append_privs: yes`.

What implementation idea would you prefer?

betanummeric:
##### SUMMARY
fixes #331

Add an argument `subtract_privs` (boolean, default false, mutual conflict with `append_privs`) to mysql_user and mysql_role. If true, the privileges specified by the `priv` argument get revoked and unspecified privileges remain unchanged.

##### ISSUE TYPE
- Feature Pull Request
##### COMPONENT NAME
mysql_user and mysql_role

### DIFF
--- a/changelogs/fragments/333-mysql_user-mysql_role-add-subtract_privileges-argument.yml
+++ b/changelogs/fragments/333-mysql_user-mysql_role-add-subtract_privileges-argument.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "mysql_user and mysql_role: Add the argument ``subtract_privs`` (boolean, default false, mutually exclusive with ``append_privs``). If set, the privileges given in ``priv`` are revoked and existing privileges are kept. (https://github.com/ansible-collections/community.mysql/pull/333)"
--- a/changelogs/fragments/333-mysql_user-mysql_role-add-subtract_privileges-argument.yml
+++ b/changelogs/fragments/333-mysql_user-mysql_role-add-subtract_privileges-argument.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - "mysql_user and mysql_role: Add the argument ``subtract_privs`` (boolean, default false, mutually exclusive with ``append_privs``). If set, the privileges given in ``priv`` are revoked and existing privileges are kept. (https://github.com/ansible-collections/community.mysql/pull/333)"
+  - "mysql_user and mysql_role: Add the argument ``subtract_privs`` (boolean, default false, mutually exclusive with ``append_privs``). If set, the privileges given in ``priv`` are revoked and existing privileges are kept (https://github.com/ansible-collections/community.mysql/pull/333)."
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -60,6 +60,7 @@ options:
       - Revoke the privileges defined by the I(priv) option and keep other existing privileges.
         If set, invalid privileges in I(priv) are ignored.
         Mutually exclusive with I(append_privs).
+    version_added: '3.2.0'
     type: bool
     default: no
 
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -1034,7 +1034,7 @@ def main():
             module.fail_json(msg=to_native(e))
 
         try:
-            priv = privileges_unpack(priv, mode)
+            priv = privileges_unpack(priv, mode, ensure_usage=not subtract_privs)
         except Exception as e:
             module.fail_json(msg='Invalid privileges string: %s' % to_native(e))
 
@@ -1063,6 +1063,8 @@ def main():
     try:
         if state == 'present':
             if not role.exists:
+                if subtract_privs:
+                    priv = None  # avoid granting unwanted privileges
                 changed = role.add(members, priv, module.check_mode, admin,
                                    set_default_role_all)
 
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -58,6 +58,7 @@ options:
   subtract_privs:
     description:
       - Revoke the privileges defined by the I(priv) option and keep other existing privileges.
+        If set, invalid privileges in I(priv) are ignored.
         Mutually exclusive with I(append_privs).
     type: bool
     default: no
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -51,7 +51,14 @@ options:
   append_privs:
     description:
       - Append the privileges defined by the I(priv) option to the existing ones
-        for this role instead of overwriting them.
+        for this role instead of overwriting them. Mutually exclusive with I(subtract_privs).
+    type: bool
+    default: no
+
+  subtract_privs:
+    description:
+      - Revoke the privileges defined by the I(priv) option and keep other existing privileges.
+        Mutually exclusive with I(append_privs).
     type: bool
     default: no
 
@@ -233,6 +240,14 @@ EXAMPLES = r'''
     name: business
     members:
     - marketing
+
+- name: Ensure the role foo does not have the DELETE privilege 
+  community.mysql.mysql_role:
+    state: present
+    name: foo
+    subtract_privs: yes
+    priv:
+      'db1.*': DELETE
 '''
 
 RETURN = '''#'''
@@ -821,9 +836,9 @@ class Role():
         return True
 
     def update(self, users, privs, check_mode=False,
-               append_privs=False, append_members=False,
-               detach_members=False, admin=False,
-               set_default_role_all=True):
+               append_privs=False, subtract_privs=False,
+               append_members=False, detach_members=False,
+               admin=False, set_default_role_all=True):
         """Update a role.
 
         Update a role if needed.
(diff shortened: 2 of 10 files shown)


betanummeric:
Hi @Andersson007, @rsicart, @Jorge-Rodriguez, @bmalynovytch,
Currently, when an invalid privilege is given along with `subtract_privs: yes`, it is just ignored because it is not found among the existing privileges. How should the module behave: ignore those or fail with an error message?

When granting a privilege, the `invalid priv string` error comes from [trying out the statement and catching the mysql/mariadb error](https://github.com/ansible-collections/community.mysql/blob/main/plugins/module_utils/user.py#L654-L657). This is not possible here, because it is never attempted to remove the invalid privilege. If an error is desired, a check needs to be added in the module.

The integration tests are currently red because they assume an error but the invalid privilege is ignored.

rsicart:
Good job, thanks!

I think ignoring and documenting this behavior would be enough.

What does the rest of the team think about that?

Andersson007:
> Good job, thanks!
> 
> I think ignoring and documenting this behavior would be enough.
> 
> What does the rest of the team think about that?

Hi all, sorry for the later reply - was on PTO, +1 to ignoring and documenting

Andersson007:
@betanummeric if you're OK with our opinion, feel free to proceed as i don't think we'll get more feedback. So please go on:)

betanummeric:
@Andersson007 I'm ok with documenting and ignoring. As I thought this would be your opinion, I already implemented that, see commit [ea48464](https://github.com/ansible-collections/community.mysql/pull/333/commits/ea48464f32bc9e22d84b72f0b3001c545a25cdfe) and the two after.

The integration tests only fail for mysql 8 at `test_mysql_user/tasks/test_privs.yml:153` (see [here](https://github.com/ansible-collections/community.mysql/runs/6075636560?check_suite_focus=true#step:8:19631)) with `"msg": "Error granting privileges, invalid priv string: "`. The privilege string is empty, it looks like somewhere in the module the privileges were lost. I'm not sure that this is related to my changes.

Apart from that, I am done with my changes.

changelogs/fragments/333-mysql_user-mysql_role-add-subtract_privileges-argument.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "mysql_user and mysql_role: Add the argument ``subtract_privs`` (boolean, default false, mutually exclusive with ``append_privs``). If set, the privileges given in ``priv`` are revoked and existing privileges are kept. (https://github.com/ansible-collections/community.mysql/pull/333)"
```
Andersson007:
```suggestion
  - "mysql_user and mysql_role: Add the argument ``subtract_privs`` (boolean, default false, mutually exclusive with ``append_privs``). If set, the privileges given in ``priv`` are revoked and existing privileges are kept (https://github.com/ansible-collections/community.mysql/pull/333)."
```
just to make all changelog entries looking similar

Andersson007:
@betanummeric i've taken a look at the changes and have realized that i need to start my day with it:)

Before that:
>I'm not sure that this is related to my changes.

I ran the tests locally without the changes and against MySQL 8 - it works and this test (that fails) had been there before this PR, so the failings are probably related to the changes.

betanummeric:
The error described in my comment above is fixed with commit [29885d1](https://github.com/ansible-collections/community.mysql/pull/333/commits/29885d12430eea8620e051550c215c84b6033179). The issue was that the difference between requested and existing privileges only was the `WITH GRANT OPTION`. The module tried to do as little as possible, which resulted in invalid SQL syntax: `GRANT ON *.* to user@host WITH GRANT OPTION` (missing privileges).
I chose to fix it by adding all existing privileges to the empty list. It would also work with adding all requested privileges, any subset of them or just `USAGE`. That should make no difference here.

betanummeric:
I have to correct myself: The placeholder privileges for only adding the grant option make a difference indeed, because privileges are first revoked, then granted. I changed the placeholder to `USAGE` and I hope the integration test succeed this time.

Finding this bug was tricky because the module is not very verbose about what it does. I added some more output in commit [37fa66a](https://github.com/ansible-collections/community.mysql/pull/333/commits/37fa66a836835da53f875980516d490da49df9e2), because I first couldn't reproduce the bug. The added output will only show the added/revoked privileges of the last db/table, it is by far not comprehensive.

Andersson007:
@betanummeric sorry for the delay, busy days.., i'll try to review it tomorrow morning, thanks for your work!

tests/integration/targets/test_mysql_user/tasks/test_privs.yml
```diff
@@ -173,7 +173,7 @@
         state: present
       register: result
 
-    # FIXME: on mariadb 10.5 there's always a change
+    # FIXME: on mariadb >=10.5.2 there's always a change because the REPLICATION CLIENT privilege was renamed to BINLOG MONITOR
```
Andersson007, line 176:
```suggestion
    # FIXME: on mariadb >=10.5.2 there's always a change because the REPLICATION CLIENT privilege was renamed to BINLOG MONITOR
```
thanks for the clarification:)

Andersson007 (commented):
@betanummeric solid refactoring, thanks:)

Also thanks for adding the comments (they make things much clearer) and good test coverage

LGTM, one formatting thing from me

I believe one day we should split all of those ancient long core functions into several smaller ones doing one thing, etc. and cover them with unit tests in future.

Let's wait for @rsicart 's review

Andersson007 (approved):

plugins/modules/mysql_role.py
```diff
@@ -51,7 +51,15 @@
...
-        for this role instead of overwriting them.
+        for this role instead of overwriting them. Mutually exclusive with I(subtract_privs).
+    type: bool
+    default: no
+
+  subtract_privs:
+    description:
+      - Revoke the privileges defined by the I(priv) option and keep other existing privileges.
+        If set, invalid privileges in I(priv) are ignored.
+        Mutually exclusive with I(append_privs).
```
Andersson007, line 62:
```suggestion
        Mutually exclusive with I(append_privs).
    version_added: '3.2.0'
```

plugins/modules/mysql_user.py
```diff
@@ -63,7 +63,14 @@
...
       - Append the privileges defined by priv to the existing ones for this
-        user instead of overwriting existing ones.
+        user instead of overwriting existing ones. Mutually exclusive with I(subtract_privs).
+    type: bool
+    default: no
+  subtract_privs:
+    description:
+      - Revoke the privileges defined by the I(priv) option and keep other existing privileges.
+        If set, invalid privileges in I(priv) are ignored.
+        Mutually exclusive with I(append_privs).
```
Andersson007, line 73:
```suggestion
        Mutually exclusive with I(append_privs).
    version_added: '3.2.0'
```

Andersson007 (commented):
@betanummeric , completely forgot about `version_added`, could you please add it to the options so that users will see since what version it is available.

rsicart:
Thanks for contributing @betanummeric !

And thanks for reviewing @Andersson007 , indeed that's a good thing to have a double review when a PR is related to critical parts of code, like this one :)

All changes seem coherent and logical to me, thanks for commenting the code, that helped a lot. Integration tests are also clear and easy to read.

I needed some time to understand why, when creating a user/role with privs and substract_privs, we set privs to None. Do you confirm that that's because:
* we can't substract privs from a non existing user/role
* otherwise that would create a user/role with privs

Good job!
LGTM :+1:

rsicart (approved):

betanummeric:
Thanks for the feedback!

@rsicart
When a user/role needs to be created and `subtract_privs: true` is set, we need to set `priv = None` because otherwise ~there would follow a `REVOKE USAGE ON *.* FROM the_just_created_user_or_role`, even if `'*.*': 'USAGE'` is not in `priv`.~ `privs` would be granted to the user/role, which is clearly against the purpose of `subtract_privs`.  I see no purpose in revoking privileges from a freshly created user/role, so editing privileges is skipped by setting `priv = None`.

Andersson007:
@betanummeric thank you for the contribution!
@rsicart thank you for reviewing!

I think we can release the collection soon. As it's a time consuming process, would be great to have that blocker https://github.com/ansible-collections/community.mysql/pull/341 merged as well. I'd be grateful for reviews.

UPDATE: there's another bugfix which is ready for review https://github.com/ansible-collections/community.mysql/pull/322

PR #352, opened 2022-05-12
DNM GitHub Action hacking

gundalow:
See if we can get ansible-test to use the git diff that GHA returns


Andersson007:
I'll close it for a while

PR #369, opened 2022-05-23
mysql_role: add argument "members_must_exist"

### LINKED ISSUE #366: mysql_role: multiple bugs with detach_members
##### ISSUE TYPE
- Bug Report

##### AFFECTED VERSIONS
ansible 2.10.8
community.mysql 3.2.1, mysql_role module

##### SUMMARY
The `detach_members` argument of the mysql_role module is documented as follows:
> Detaches members defined by the members option from the role instead of overwriting all the current members.

### adding unwanted members
When the role does not exist, it is created and the users specified in `members` are added instead of detached, regardless of `detach_members`.

See the lack of `detach_members` here: https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_role.py#L1067-L1071

### detaching too few members
Detecting the existing members of a role does not work if the `SHOW GRANTS` statement uses other quotes than backticks (&#96;). I checked MariaDB 10.{1-6} and MySQL 5.7, 8.0, with and without ANSI_QUOTES SQL-mode, and found either single or double quotes (`'`/`"`), so I don't know why the module expects backticks. This causes that existing members are added again (redundant) and no unwanted members are detached.

The backticks are assumed here:
https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_role.py#L691
https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_role.py#L697

On MariaDB 10.6, notice the double quotes:
```
mysql ((none))> create user testuser;
Query OK, 0 rows affected (0.005 sec)

mysql ((none))> grant testrole1 to testuser;
Query OK, 0 rows affected (0.003 sec)

mysql ((none))> show grants for testuser;
+--------------------------------------+
| Grants for testuser@%                |
+--------------------------------------+
| GRANT "testrole1" TO "testuser"@"%"  |
| GRANT USAGE ON *.* TO "testuser"@"%" |
+--------------------------------------+
2 rows in set (0.001 sec)

mysql ((none))> select version();
+--------------------------------------------+
| version()                                  |
+--------------------------------------------+
| 10.6.4-MariaDB-1:10.6.4+maria~bullseye-log |
+--------------------------------------------+
1 row in set (0.001 sec)
```

### raising too much errors
When a user which shall be detached from a role does not exist, the module fails. I would like the module to ignore this user. Other people may expect the failure, so we could add a flag (e.g. `members_must_exist`, default true) to toggle failing/ignoring.

#### steps to reproduce
```yaml
- name: ensure the user does not exist
  community.mysql.mysql_user:
    login_unix_socket: /tmp/mysql-3306.sock
    name: missing_user
    host: localhost
    state: absent
- name: ensure the role does not exist
  community.mysql.mysql_role:
    login_unix_socket: /tmp/mysql-3306.sock
    state: absent
    name: testrole1
- name: create role  # this will fail
  community.mysql.mysql_role:
    login_unix_socket: /tmp/mysql-3306.sock
    state: present
    name: testrole1
    detach_members: yes
    members:
      - missing_user@localhost
```

betanummeric:
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
mysql_role

##### SUMMARY
Add a new argument "members_must_exist" (boolean, default true).

The assertion that the users supplied in the "members" argument exist is only executed when the new argument "members_must_exist" is true, to allow opt-out.

fixes #366 (the "raising too much errors" part)

### DIFF
--- a/changelogs/fragments/369_mysql_role-add-members_must_exist.yml
+++ b/changelogs/fragments/369_mysql_role-add-members_must_exist.yml
@@ -1,5 +1,4 @@
 minor_changes:
   - >
-    mysql_role: Add the argument "members_must_exist" (boolean, default true). The assertion that the users supplied in
-    the "members" argument exist is only executed when the new argument "members_must_exist" is true, to allow opt-out.
-    See (https://github.com/ansible-collections/community.mysql/pull/369).
+    mysql_role - add the argument ``members_must_exist`` (boolean, default true). The assertion that the users supplied in
+    the ``members`` argument exist is only executed when the new argument ``members_must_exist`` is ``true``, to allow opt-out (https://github.com/ansible-collections/community.mysql/pull/369).
--- a/changelogs/fragments/369_mysql_role-add-members_must_exist.yml
+++ b/changelogs/fragments/369_mysql_role-add-members_must_exist.yml
@@ -1,0 +1,5 @@
+minor_changes:
+  - >
+    mysql_role: Add the argument "members_must_exist" (boolean, default true). The assertion that the users supplied in
+    the "members" argument exist is only executed when the new argument "members_must_exist" is true, to allow opt-out.
+    See (https://github.com/ansible-collections/community.mysql/pull/369).
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -389,6 +389,11 @@ class DbServer():
                 msg = 'User / role `%s` with host `%s` does not exist' % (user[0], user[1])
                 self.module.fail_json(msg=msg)
 
+    def filter_existing_users(self, users):
+        for user in users:
+            if user in self.users:
+                yield user
+
     def __get_users(self):
         """Get users.
 
@@ -1028,8 +1033,10 @@ def main():
 
     if members:
         members = normalize_users(module, members, server.is_mariadb())
-        if members_must_exist and not detach_members:
+        if members_must_exist:
             server.check_users_in_db(members)
+        else:
+            members = list(server.filter_existing_users(members))
 
     # Main job starts here
     role = Role(module, cursor, name, server)
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -114,6 +114,13 @@ options:
     type: bool
     default: no
 
+  members_must_exist:
+    description:
+      - When C(yes), the module fails if any user in C(members) does not exist.
+      - When C(no), users in C(members) which don't exist are simply skipped.
+    type: bool
+    default: yes
+
 notes:
   - Pay attention that the module runs C(SET DEFAULT ROLE ALL TO)
     all the I(members) passed by default when the state has changed.
@@ -918,6 +925,7 @@ def main():
         detach_members=dict(type='bool', default=False),
         check_implicit_admin=dict(type='bool', default=False),
         set_default_role_all=dict(type='bool', default=True),
+        members_must_exist=dict(type='bool', default=True)
     )
     module = AnsibleModule(
         argument_spec=argument_spec,
@@ -951,6 +959,7 @@ def main():
     check_hostname = module.params['check_hostname']
     db = ''
     set_default_role_all = module.params['set_default_role_all']
+    members_must_exist = module.params['members_must_exist']
 
     if priv and not isinstance(priv, (str, dict)):
         msg = ('The "priv" parameter must be str or dict '
@@ -1019,7 +1028,8 @@ def main():
 
     if members:
         members = normalize_users(module, members, server.is_mariadb())
-        server.check_users_in_db(members)
+        if members_must_exist and not detach_members:
+            server.check_users_in_db(members)
 
     # Main job starts here
     role = Role(module, cursor, name, server)
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -116,8 +116,8 @@ options:
 
   members_must_exist:
     description:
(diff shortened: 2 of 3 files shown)


plugins/modules/mysql_role.py
```diff
@@ -114,6 +114,13 @@
     type: bool
     default: no
 
+  members_must_exist:
+    description:
+      - When C(yes), the module fails if any user in C(members) does not exist.
+      - When C(no), users in C(members) which don't exist are simply skipped.
```
Andersson007:
```suggestion
      - When C(yes), the module fails if any user in I(members) does not exist.
      - When C(no), users in I(members) which don't exist are simply skipped.
```

plugins/modules/mysql_role.py
```diff
@@ -114,6 +114,13 @@
     type: bool
     default: no
 
+  members_must_exist:
```
Andersson007, line 117:
```suggestion
  fail_on_member:
```
Maybe? It's just an option from me. Let's also ask @bmalynovytch @rsicart @hubiongithub and @Jorge-Rodriguez  about naming

changelogs/fragments/369_mysql_role-add-members_must_exist.yml
```diff
@@ -0,0 +1,5 @@
+minor_changes:
+  - >
+    mysql_role: Add the argument "members_must_exist" (boolean, default true). The assertion that the users supplied in
+    the "members" argument exist is only executed when the new argument "members_must_exist" is true, to allow opt-out.
+    See (https://github.com/ansible-collections/community.mysql/pull/369).
```
Andersson007:
```suggestion
    mysql_role - add the argument ``members_must_exist`` (boolean, default true). The assertion that the users supplied in
    the ``members`` argument exist is only executed when the new argument ``members_must_exist`` is ``true``, to allow opt-out (https://github.com/ansible-collections/community.mysql/pull/369).
```
To make entries consistent

Andersson007 (commented):
@betanummeric thanks very much for fixing and improving the module! I like the idea.
Speaking about the default value, we can't merge it now if the it's `yes` as it would be a breaking change (we are following SemVer and such changes must be announced in advance with warnings, etc., and released in a major release).

I suggest making it `no` for now. We could discuss and change it later in a separate issue. If the community will be OK, we'll create a PR which we'll merge before the next major release (which i don't think will happen earlier than in half a year).

So to proceed now, I suggest making it `no` by default (for now).

betanummeric:
@Andersson007 The current default `members_must_exist: yes` should preserve the behavior, so this should not be a breaking change. I previously had the variable named `ignore_missing_members` with the value flipped (default `no`), but I thought it was an unnecessary negation and renamed it to `members_must_exist`.

Andersson007:
> @Andersson007 The current default `members_must_exist: yes` should preserve the behavior, so this should not be a breaking change. I previously had the variable named `ignore_missing_members` with the value flipped (default `no`), but I thought it was an unnecessary negation and renamed it to `members_must_exist`.

Ah, got it now, busy days... sorry:)
OK, let's wait for folks' opinions for a day or two (`fail_on_member` came to mind because there's a (not fully similar) parameter called `fail_on_user` in the postgresql_user module).
Could you also please resolve the conflict (https://docs.ansible.com/ansible/latest/dev_guide/developing_rebasing.html use `main` instead of `devel`)

Andersson007:
@betanummeric there's an integration test error in case you've missed it

plugins/modules/mysql_role.py
rsicart: Both names are well chosen, I like both. Using `fail_on_member` would be more uniform with postgresql, but `members_must_exist` sounds more explicit to me.

Andersson007: cool, let's proceed with `members_must_exist` then:) thanks

betanummeric:
@Andersson007  Thanks for the hint, but it looks like the error is unrelated to my changes:
```
Run pip install https://github.com/ansible/ansible/archive/stable-2.13.tar.gz --disable-pip-version-check
Collecting https://github.com/ansible/ansible/archive/stable-2.13.tar.gz
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='github.com', port=443): Read timed out. (read timeout=15)")': /ansible/ansible/archive/stable-2.13.tar.gz
  ERROR: HTTP error 502 while getting https://github.com/ansible/ansible/archive/stable-2.13.tar.gz
ERROR: Could not install requirement https://github.com/ansible/ansible/archive/stable-2.13.tar.gz because of HTTP error 502 Server Error: Bad Gateway for url: https://github.com/ansible/ansible/archive/stable-2.13.tar.gz for URL https://github.com/ansible/ansible/archive/stable-2.13.tar.gz
```
All other integration tests ran successfully. Can you maybe reschedule the failed test?

Andersson007:
@betanummeric ah, OK, thanks. Unfortunately there's no way to re-launch only one task in GHA, so i closed reopened the PR to trigger another run

Andersson007:
ah, GitHub has issues now https://www.githubstatus.com/. I'll relauch the tests later

Andersson007 (approved):
LGTM

Andersson007:
@betanummeric thanks for the improvement!
@rsicart thanks for reviewing!

Andersson007:
@betanummeric ah, should we maybe add an example of the new param use to the EXAMPLES block? Would you like to submit a PR? (no changelog fragment needed)

PR #400, opened 2022-06-14
Note added regarding the default config file, ~/.my.cnf

chriscroome:
Update the notes to reflect the module behaviour regarding `~/.my.cnf` files that don't contain `[client]` section.

### DIFF
--- a/plugins/doc_fragments/mysql.py
+++ b/plugins/doc_fragments/mysql.py
@@ -98,4 +98,9 @@ notes:
    - Alternatively, you can use the mysqlclient library instead of MySQL-python (MySQLdb)
      which supports both Python 2.X and Python >=3.5.
      See U(https://pypi.org/project/mysqlclient/) how to install it.
+   - The config file, C(~/.my.cnf), if it exists, must contain a C([client]) section as a MySQL connector requirement.
+   - The default config file, C(~/.my.cnf) will be read, if it exists, even if I(config_file) is not specified.
+   - To prevent the default config file from being read, set I(config_file) to be an empty string.
+   - If credentials from the config file (for example, C(/root/.my.cnf)) are not needed to connect to a database server but
+     the file exists and does not contain a C([client]) section, it will be read and will cause the connection to fail.
 '''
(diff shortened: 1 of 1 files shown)


plugins/doc_fragments/mysql.py
```diff
@@ -98,4 +98,9 @@ class ModuleDocFragment(object):
    - Alternatively, you can use the mysqlclient library instead of MySQL-python (MySQLdb)
      which supports both Python 2.X and Python >=3.5.
      See U(https://pypi.org/project/mysqlclient/) how to install it.
+   - The config file, C(~/.my.cnf), if it exists, must contain a C([client]) section as a MySQL connector requirement.
```
Andersson007:
```suggestion
   - The config file, C(~/.my.cnf), if it exists, must contain a C([client]) section as a MySQL connector requirement.
```
Should we also add this particular sentence to the `config_file` option description?

chriscroome: Good idea, I also wonder if two other sentences should also be moved there, so that part could have four notes in total:

- Specify a config file from which user and password are to be read.
- The default config file, `~/.my.cnf` will be read, if it exists, even if *config_file* is not specified.
- The config file, if it exists, must contain a `[client]` section as a MySQL connector requirement.
- To prevent the default config file from being read, set *config_file* to be an empty string.

And then just this one added at the end?

- If credentials from the `config_file` (for example, `/root/.my.cnf`) are not needed to connect to a database server, but the file exists and does not contain a `[client]` section, before any other valid directives, it will be read and this will cause the connection to fail, to prevent this set it to an empty string, (for example `config_file: ''`).

plugins/doc_fragments/mysql.py
```diff
@@ -98,4 +101,7 @@ class ModuleDocFragment(object):
    - Alternatively, you can use the mysqlclient library instead of MySQL-python (MySQLdb)
      which supports both Python 2.X and Python >=3.5.
      See U(https://pypi.org/project/mysqlclient/) how to install it.
+   - If credentials from the config file (for example, C(/root/.my.cnf)) are not needed to connect to a database server, but
+     the file exists and does not contain a C([client]) section, before any other valid directives, it will be read and this
+     will cause the connection to fail, to prevent this set it to an empty string, (for example C(config_file: '')).
```
Andersson007:
```suggestion
   - "If credentials from the config file (for example, C(/root/.my.cnf)) are not needed to connect to a database server, but
     the file exists and does not contain a C([client]) section, before any other valid directives, it will be read and this
     will cause the connection to fail, to prevent this set it to an empty string, (for example C(config_file: ''))."
```
The semicolon causes doc build errors, we should use double quotes in such cases.

Andersson007: @chriscroome ^

PR #403, opened 2022-06-21
Fix mysqldump ignoring errors

### LINKED ISSUE #256: mysql_db module does not detect `mysqldump` failure
##### SUMMARY
`mysql_db` module does not detect `mysqldump` failure when output is set to be a compressed file. 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`mysql_db`

##### ANSIBLE VERSION

```
ansible 2.10.15
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.10 (default, May  6 2021, 00:05:59) [GCC 10.2.1 20201203]

```

##### COLLECTION VERSION

```
Collection        Version
----------------- -------
community.general 1.3.6

```

##### OS / ENVIRONMENT

Encountered on `Debian GNU/Linux 9.13 (stretch)`.

##### STEPS TO REPRODUCE

You need to be able to get mysqldump to fail, in our case it was failing because records dumped exceeded `max_allowed_packet` default of 16MB which we verified by running `mysqldump` from CLI instead of via `mysql_db`. Perhaps setting invalid credentials may break it too.

E.g.:
```yaml
- name: backup the database
  mysql_db:
    name: mydatabase
    state: dump
    target: /tmp/backup.gz
    login_host: ~
    login_port: ~
    login_user: ~
    login_password: ~
    dump_extra_args: "--max_allowed_packet=1MB"
```

##### EXPECTED RESULTS

Task fails as the value of `max_allowed_packet` doesn't permit it to successfully complete `mysqldump` when dumping an object bigger than 1MB.

##### ACTUAL RESULTS

Task completes successfully with `/tmp/backup.gz` file created which contains valid SQL file. However the backup is incomplete - i.e. missing data and/or structure as the `mysqldump` process exited early with an error.

I believe this is because the output is set to be gzip and ansible pipes the output from `mysqldump` into `gzip`, however unless `set -o pipefail` is set for that shell's environment, failure of `mysqldump` won't be noticed and a partial mysql dump file will be gzipped.

I regard this as a bug since it's expected that if the module provides automatic packaging, it should handle the dump failure regarding of what the default shell settings may be.

See https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_db.py#L420 for the relevant code.

laurent-indermuehle:
##### SUMMARY
The mysql_db module does not display mysqldump errors when compression is used.

Fixes #256 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

mysql_db

##### ADDITIONAL INFORMATION

The issue #256 initially was met the first time because of a too big field and a `max_allowed_packet` to small. As this is difficult to test, the tests were written with a MySQL view pointing to a non-existent table.

```paste below
# Before
changed: [testhost]

# After
fatal: [testhost]: FAILED! => {"changed": false, "msg": "mysqldump: [Warning] Using a password on the command line interface can be insecure.\nmysqldump: Got error: 1356: View 'db2.v1' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them when using LOCK TABLES\n"}
```

### DIFF
--- a/changelogs/fragments/fix-256-mysql_dump-errors.yml
+++ b/changelogs/fragments/fix-256-mysql_dump-errors.yml
@@ -4,4 +4,4 @@ bugfixes:
   - mysql_dump - Fixes issue 256. Using compression masks errors messages from
     mysql_dump. By default the fix is inactiv to ensure retro-compatibility
     with system without bash. To activate the fix, use the module option
-    `pipefail=true`.
+    ``pipefail=true`` (https://github.com/ansible-collections/community.mysql/issues/256).
--- a/changelogs/fragments/fix-256-mysql_dump-errors.yml
+++ b/changelogs/fragments/fix-256-mysql_dump-errors.yml
@@ -1,0 +1,7 @@
+---
+
+bugfixes:
+  - mysql_dump - Fixes issue 256. Using compression masks errors messages from
+    mysql_dump. By default the fix is inactiv to ensure retro-compatibility
+    with system without bash. To activate the fix, use the module option
+    `pipefail=true`.
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -156,6 +156,14 @@ options:
     - Can be useful, for example, when I(state=import) and a dump file contains relative paths.
     type: path
     version_added: '3.4.0'
+  pipefail:
+    description:
+    - Use bash instead of sh and add -o pipefail to catch errors from the
+      mysql_dump command when compression is used. The default is false to
+      prevent issue on system without bash. The default may change in a future release
+    type: bool
+    default: no
+    version_added: '3.3.1'
 
 seealso:
 - module: community.mysql.mysql_info
@@ -355,7 +363,7 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
             single_transaction=None, quick=None, ignore_tables=None, hex_blob=None,
             encoding=None, force=False, master_data=0, skip_lock_tables=False,
             dump_extra_args=None, unsafe_password=False, restrict_config_file=False,
-            check_implicit_admin=False):
+            check_implicit_admin=False, pipefail=False):
     cmd = module.get_bin_path('mysqldump', True)
     # If defined, mysqldump demands --defaults-extra-file be the first option
     if config_file:
@@ -422,13 +430,20 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
     elif os.path.splitext(target)[-1] == '.xz':
         path = module.get_bin_path('xz', True)
 
-    if path:
+    if pipefail:
         cmd = 'set -o pipefail && %s | %s > %s' % (cmd, path, shlex_quote(target))
+    elif path:
+        cmd = '%s | %s > %s' % (cmd, path, shlex_quote(target))
     else:
         cmd += " > %s" % shlex_quote(target)
 
     executed_commands.append(cmd)
-    rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True, executable='bash')
+
+    if pipefail:
+      rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True, executable='bash')
+    else:
+      rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True)
+
     return rc, stdout, stderr
 
 
@@ -569,6 +584,7 @@ def main():
         check_implicit_admin=dict(type='bool', default=False),
         config_overrides_defaults=dict(type='bool', default=False),
         chdir=dict(type='path'),
+        pipefail=dict(type='bool', default=False),
     )
 
     module = AnsibleModule(
@@ -618,6 +634,7 @@ def main():
     check_implicit_admin = module.params['check_implicit_admin']
     config_overrides_defaults = module.params['config_overrides_defaults']
     chdir = module.params['chdir']
+    pipefail = module.params['pipefail']
 
     if chdir:
         try:
@@ -704,7 +721,7 @@ def main():
                                      ssl_ca, single_transaction, quick, ignore_tables,
                                      hex_blob, encoding, force, master_data, skip_lock_tables,
                                      dump_extra_args, unsafe_login_password, restrict_config_file,
-                                     check_implicit_admin)
+                                     check_implicit_admin, pipefail)
         if rc != 0:
             module.fail_json(msg="%s" % stderr)
         module.exit_json(changed=True, db=db_name, db_list=db, msg=stdout,
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -430,7 +430,7 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
     elif os.path.splitext(target)[-1] == '.xz':
         path = module.get_bin_path('xz', True)
 
-    if pipefail:
+    if pipefail and path:
         cmd = 'set -o pipefail && %s | %s > %s' % (cmd, path, shlex_quote(target))
     elif path:
         cmd = '%s | %s > %s' % (cmd, path, shlex_quote(target))
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -423,12 +423,12 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
         path = module.get_bin_path('xz', True)
 
     if path:
-        cmd = '%s | %s > %s' % (cmd, path, shlex_quote(target))
+        cmd = 'set -o pipefail && %s | %s > %s' % (cmd, path, shlex_quote(target))
     else:
         cmd += " > %s" % shlex_quote(target)
 
     executed_commands.append(cmd)
-    rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True)
+    rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True, executable='bash')
     return rc, stdout, stderr
 
 
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -437,10 +437,10 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
     elif os.path.splitext(target)[-1] == '.xz':
         path = module.get_bin_path('xz', True)
 
-    if pipefail and path:
-        cmd = 'set -o pipefail && %s | %s > %s' % (cmd, path, shlex_quote(target))
-    elif path:
+    if path:
         cmd = '%s | %s > %s' % (cmd, path, shlex_quote(target))
+        if pipefail:
+            cmd = 'set -o pipefail && ' + cmd
     else:
         cmd += " > %s" % shlex_quote(target)
 
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -440,9 +440,9 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
     executed_commands.append(cmd)
 
     if pipefail:
-      rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True, executable='bash')
+        rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True, executable='bash')
     else:
-      rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True)
+        rc, stdout, stderr = module.run_command(cmd, use_unsafe_shell=True)
 
     return rc, stdout, stderr
 
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -159,11 +159,11 @@ options:
   pipefail:
     description:
     - Use C(bash) instead of C(sh) and add C(-o pipefail) to catch errors from the
-      mysql_dump command when compression is used. The default is false to
-      prevent issue on system without bash. The default may change in a future release
+      mysql_dump command when I(state=import) and compression is used. The default is I(false) to
+      prevent issue on system without bash. The default may change in a future release.
     type: bool
     default: no
-    version_added: '3.3.1'
+    version_added: '3.4.0'
 
 seealso:
 - module: community.mysql.mysql_info
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -303,6 +303,13 @@ EXAMPLES = r'''
     login_password: 123456
     name: bobdata
     state: present
+
+- name: Dump a database with compression and catch errors from mysqldump with bash pipefail
+  community.mysql.mysql_db:
+    state: dump
+    name: foo
+    target: /tmp/dump.sql.gz
+    pipefail: true
 '''
 
 RETURN = r'''
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -158,7 +158,7 @@ options:
     version_added: '3.4.0'
   pipefail:
     description:
-    - Use bash instead of sh and add -o pipefail to catch errors from the
+    - Use C(bash) instead of C(sh) and add C(-o pipefail) to catch errors from the
       mysql_dump command when compression is used. The default is false to
       prevent issue on system without bash. The default may change in a future release
     type: bool
--- a/tests/integration/targets/test_mysql_db/tasks/state_present_absent.yml
+++ b/tests/integration/targets/test_mysql_db/tasks/state_present_absent.yml
@@ -18,8 +18,8 @@
 # ============================================================
 - name: remove database if it exists
   command: >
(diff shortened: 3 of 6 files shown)


Andersson007:
@laurent-indermuehle thanks for the fix!

The CI tests have failed. Please see the logs.

Anyway I'm afraid in this implementation it'll be a breaking change.. at least for FreeBSD users as `bash` is not installed there by default (I read it in the [official doc](https://docs.freebsd.org/en/articles/linux-users/#:~:text=Linux%C2%AE%20users%20are%20often,as%20the%20default%20user%20shell.))
Any ideas on how to overcome it?
Maybe a new boolean option that will allow to do what you added with `no` by default for now to preserve backwards compatibility in systems with no bash?
Then we could properly change the default later in the next major release.

laurent-indermuehle:
@Andersson007 I didn't thought that bash could be missing on some systems.
I like your idea of a quick fix using the bool with no by default. I'll try to implement that.

For the CI, it's my first time working with Github Actions. I see that some tests are returning a none 0 code but have no clue why. On my machine, units, integrations and sanity tests passes.

Andersson007:
> @Andersson007 I didn't thought that bash could be missing on some systems. I like your idea of a quick fix using the bool with no by default. I'll try to implement that.

Thanks!

> For the CI, it's my first time working with Github Actions. I see that some tests are returning a none 0 code but have no clue why. On my machine, units, integrations and sanity tests passes.

It's because your local fork contains default versions of things we test against (listed in the [setup_mysql](https://github.com/ansible-collections/community.mysql/blob/main/tests/integration/targets/setup_mysql/defaults/main.yml) role that prepares the test environment and is run before every target as a dependency). 
There's a file `https://github.com/ansible-collections/community.mysql/blob/main/.github/workflows/ansible-test-plugins.yml`.
It changes the defaults in the role.

When you're debugging, the easiest (imo) algorithm is to:
1. click `Details` near a failing test
2. click the gear in the upper-right corener
3. choose `View raw logs`
4. use search in browser with `FAILED`, ignore things marked with `ignoring`

Following these steps, i can see that the error is
```
2022-06-22T05:52:42.2627776Z TASK [test_mysql_db : Check dumps errors | Setup test | Create 2 schemas] ******
2022-06-22T05:52:42.6390493Z [0;31mfatal: [testhost]: FAILED! => {"changed": false, "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: (2003, \"Can't connect to MySQL server on 'localhost' ([Errno 99] Cannot assign requested address)\")"}[0m
2022-06-22T05:52:42.7416213Z 
```

Andersson007:
It looks like only tests against Python 3.6 failed

laurent-indermuehle:
@Andersson007 Thanks for the workflow to read GH Actions logs! It helps :)

I'm not sure what the issue is. I see so many things that looks weird to me:

1) True instead of true or yes: https://github.com/ansible-collections/community.mysql/blob/main/tests/integration/targets/test_mysql_db/tasks/state_present_absent.yml#L22

2) No such file or directory on Python 3.6:
```bash
TASK [test_mysql_db : remove database if it exists] 
[fatal: [testhost]: FAILED! => {
  "cmd": "'mysql -uroot -pmsandbox -P3307 --protocol=tcp -sse '\"'\"'drop database data'\"'\"''",
  "msg": "[Errno 2] No such file or directory: b\"mysql -uroot -pmsandbox -P3307 --protocol=tcp -sse 'drop database data'\": b\"mysql -uroot -pmsandbox -P3307 --protocol=tcp -sse 'drop database data'\"",
  "rc": 2}
```

3) Unable to drop 'data' schema:
```bash
TASK [test_mysql_db : make sure the test database is not there] 
[changed: [testhost] => {
  "changed": true, 
  "cmd": ["mysql", "-uroot", "-pmsandbox", "-P3307", "--protocol=tcp", "data"],
  "msg": "non-zero return code", "rc": 1,
  "stderr": "mysql: ERROR 1049 (42000): Unknown database 'data'", 
}
```

I'm running ansible-test locally for Python 3.6. I'll see if the problem is the same on my machine.

laurent-indermuehle:
Ok, so MySQL8 causes issue with its new authentication plugin:
```
TASK [test_mysql_db : state dump/import - create database]
unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. 
Exception message: cryptography is required for sha256_password or caching_sha2_password
```

I'm starting to wounder if using module_defaults was a good idea? https://github.com/laurent-indermuehle/community.mysql/blob/fix_mysqldump_ignore_errors/tests/integration/targets/test_mysql_db/tasks/issue_256_mysqldump_errors.yml#L6-L12

laurent-indermuehle:
ok, I missed the blue-green 'ignoring' ;)
Many errors are the actuals tests... Searching in the row log removes the colors :P

laurent-indermuehle:
I pushed a version with the new option "pipefail".
The tests are enhanced and continue to pass. I hope the CI will be happy thought.

Tests continues to fails using mysql 8 due to a missing package cryptography.

I added a changelog fragment and a documentation for the option. I put the "version_added" at 3.3.1 but not sure what number to put.

Do I have forgot anything?

changelogs/fragments/fix-256-mysql_dump-errors.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+bugfixes:
+  - mysql_dump - Fixes issue 256. Using compression masks errors messages from
+    mysql_dump. By default the fix is inactiv to ensure retro-compatibility
+    with system without bash. To activate the fix, use the module option
+    `pipefail=true`.
```
Andersson007:
```suggestion
    ``pipefail=true`` (https://github.com/ansible-collections/community.mysql/issues/256).
```
(the changelog will be in rst format)

plugins/modules/mysql_db.py
```diff
@@ -156,6 +156,14 @@
     - Can be useful, for example, when I(state=import) and a dump file contains relative paths.
     type: path
     version_added: '3.4.0'
+  pipefail:
+    description:
+    - Use bash instead of sh and add -o pipefail to catch errors from the
```
Andersson007:
```suggestion
    - Use C(bash) instead of C(sh) and add C(-o pipefail) to catch errors from the
```

plugins/modules/mysql_db.py
```diff
@@ -156,6 +156,14 @@
...
+      mysql_dump command when compression is used. The default is false to
+      prevent issue on system without bash. The default may change in a future release
+    type: bool
+    default: no
+    version_added: '3.3.1'
```
Andersson007:
```suggestion
      mysql_dump command when I(state=import) and compression is used. The default is I(false) to
      prevent issue on system without bash. The default may change in a future release.
    type: bool
    default: no
    version_added: '3.4.0'
```

plugins/modules/mysql_db.py
```diff
@@ -422,13 +430,20 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
     elif os.path.splitext(target)[-1] == '.xz':
         path = module.get_bin_path('xz', True)
 
-    if path:
+    if pipefail:
+        cmd = 'set -o pipefail && %s | %s > %s' % (cmd, path, shlex_quote(target))
+    elif path:
```
Andersson007:
```suggestion
    if pipefail:
        cmd = 'set -o pipefail && %s | %s > %s' % (cmd, path, shlex_quote(target))
    elif path:
```
are they mutually exclusive with `path`?

Andersson007 (commented):
@laurent-indermuehle thanks!

I think it would be nice to add an example to the EXAMPLES block.

I took only a quick look today. I'll take a deeper one tomorrow or on Friday. Thanks!

plugins/modules/mysql_db.py
laurent-indermuehle: uh, you're right, I should have written:
```python
if pipefail and path:
  cmd = '...'
elif path:
  cmd = '...'
 else:
   cmd += '...'
```

laurent-indermuehle:
@Andersson007 I'm done for today. I'm sorry for the number of commits. I'm not used to work on Ansible plugins... yet ;)

Thank you if you can review the tests and fixes.
I took the liberty to fix some stuffs related to others tests (IF EXISTS, python3-cryptography), I hope that's ok.

Andersson007:
> @Andersson007 I'm done for today. I'm sorry for the number of commits. I'm not used to work on Ansible plugins... yet ;)

@laurent-indermuehle no problem at all! Please do any number of commits/ask any questions needed! Thanks for working on this:)

plugins/modules/mysql_db.py
```diff
@@ -422,13 +430,20 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
     elif os.path.splitext(target)[-1] == '.xz':
         path = module.get_bin_path('xz', True)
 
-    if path:
+    if pipefail and path:
+        cmd = 'set -o pipefail && %s | %s > %s' % (cmd, path, shlex_quote(target))
+    elif path:
         cmd = '%s | %s > %s' % (cmd, path, shlex_quote(target))
```
Andersson007:
```suggestion
    if path:
        cmd = '%s | %s > %s' % (cmd, path, shlex_quote(target))
        if pipefail:
            cmd = 'set -o pipefail && ' + cmd
```
how about this? As path is true in both the cases, we could make it a bit less verbose

Andersson007 (commented):
@laurent-indermuehle one small thing from me besides adding the example as mentioned in my yesterday's comment.

plugins/modules/mysql_db.py
laurent-indermuehle: I think your solution is better. I try to avoid nested if but here it make sense because it lessen repetitions.

One question thought: What is the minimum Python version supported? If it is 3.6+ then I propose to convert line 434 to fstring.

laurent-indermuehle: @Andersson007 Why this conversation is marked resolved?
Does that means that I can go ahead and convert to fstring? :)

Andersson007: @laurent-indermuehle ah, sorry, i thought you committed it:) Now i see. I think practically it can work with older versions, so if there's no strong reason to use incompatible things, i would avoid this.

laurent-indermuehle: @Andersson007 Ok, then no fstring.

laurent-indermuehle:
Great, I commited your changes. I think we are done with this PR, unless the CI tells otherwise.

I'm not 100% satisfied because, the default is highly insecure or prevent us from using compression. If a way exists to pipe commands together with Python subcommand.Popen() while preserving the return code of each command, it would be ways better. But I failed to find it. Especially in our case, where this method is wrapped by Ansible module.run_command().

Without help from an expert I can't offer a better solution.

I was happy to help, I learned alot and your guidances was very appreciated.

Andersson007:
> I'm not 100% satisfied because, the default is highly insecure or prevent us from using compression.

Agreed. We can change the default later in a major release

> I was happy to help, I learned alot and your guidances was very appreciated.

I'm happy to help:)

laurent-indermuehle:
After having read the documentation for the subprocess module, it seems we can use 2 Popen object and check the returned code of each.

While experimenting with that, I found the following code from mysql_db.py into the db_import() function: https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_db.py#L492-L512

I'm a bit worried about the "'Broken pipe' errors that occasionally occur" message, but aside from that, it seems to be the solution I was looking for? Right?

Andersson007:
> After having read the documentation for the subprocess module, it seems we can use 2 Popen object and check the returned code of each.
> 
> While experimenting with that, I found the following code from mysql_db.py into the db_import() function: https://github.com/ansible-collections/community.mysql/blob/main/plugins/modules/mysql_db.py#L492-L512

ah, forgot that the module uses the subprocess module directly

> I'm a bit worried about the "'Broken pipe' errors that occasionally occur" message

Yep, sounds scary

>  but aside from that, it seems to be the solution I was looking for? Right?

Would you like to continue experimenting or we should merge the PR?
Of course it would be great if everything goes seamlessly without any additional options but the current solution feels (not ideal but) safer (especially considering the "Broken pipe" comment - yeah, i recalled there was the issue, then we added the use_shell to workaround it)

I'm not sure, let's ask the other maintainers, @bmalynovytch ideas?

laurent-indermuehle:
I created a playbook and a container to test locally with a 100MB database to see if the changes affect performances.

Then I created yet another option to mysql_db called 'popen' that uses the following code to compress the dump:

```python
import gzip
from shlex import split as shlex_split
if popen:
        p1 = subprocess.Popen(shlex_split(cmd), stdout=subprocess.PIPE, stderr=subprocess.PIPE)

        with gzip.open(shlex_quote(target), 'wb', compresslevel=5) as f:
            stdout, stderr = p1.communicate()
            f.write(stdout)
            p1.wait()

        if p1.returncode != 0:
            return p1.returncode, '', to_native(stderr)
        else:
            return 0, 'Dump done', ''
```

This works, the error is captured popen the same way pipefail did. But:
- I don't know if we can add Python gzip module as a dependency
- How to handle others compressions. But aside for xz, they all seems to be available in the standard library: https://docs.python.org/3/library/archiving.html

The good news is, that the performances are almost equivalent :
  - dump sql : 2.1s
  - dump gzip : 2.47s
  - dump gzip pipefail : 2.53s
  - dump gzip popen + gzip module : 2.9s

Without the , compresslevel=5, Python defaults to 9. Which resulted in 6+ seconds per dump. I found that 5 produce the same file size as the shell version.

I'm running out of time. I'll get back in 2 weeks. What is left to do:

- Try to replace subprocess.Popen by module.run_command as it is recommended by Ansible best practices
- Try to replace gzip module by chaining the output of p1 on a second subprocess, E.G.: 
   
    ```python
    compress_cmd = shlex_split(path + ' > ' + shlex_quote(target))
    p2 = module.run_command(stdin=p1.stdout, ...)
    ```

- Try to debug broken pipe in db_import
- Align compression code between db_dump and db_import
- Find out how to use shlex.split from Ansible

laurent-indermuehle:
Here is a version without the gzip module: 
```python
from shlex import split as shlex_split
[...]
    if popen:
        compress_cmd = [module.get_bin_path('gzip', True), ' > ', shlex_quote(target)]
        p1 = subprocess.Popen(shlex_split(cmd), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
        p2 = subprocess.Popen(compress_cmd, stdin=p1.stdout, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
        stdout1, stderr1 = p1.communicate()
        stdout2, stderr2 = p2.communicate()

        if p1.returncode != 0:
            return p1.returncode, '', to_native(stderr1)
        else:
            return p2.returncode, stdout2, to_native(stderr2)
```

Should I go this route? 

I tried to use run_command(), but because it uses the .close() method, it will be impossible to pipe two run_command() together. I tried `pass_fds`, `data`, `close_fds`, with no luck. I got a broken pipe every time.

Andersson007:
@laurent-indermuehle thanks for the experimenting!

1. About the dependencies introduction  - I would avoid it but it's only my view.

2. About using Popen - IIRC that bug with the crashing pipe was floating, so the solution we already have in this PR feels like the safest.

@bmalynovytch @rsicart any ideas?

tests/integration/targets/test_mysql_db/tasks/state_present_absent.yml
```diff
@@ -18,8 +18,8 @@
 # ============================================================
 - name: remove database if it exists
   command: >
-    "{{ mysql_command }} -sse 'drop database {{ db_name }}'"
-  ignore_errors: True
+    "{{ mysql_command }} -sse 'DROP DATABASE IF EXISTS {{ db_name }}'"
+  ignore_errors: true
```
bmalynovytch, line 23: Is this related to the issue ? 🤨

bmalynovytch (approved):
The approach is the one I would have had.
Good job !

tests/integration/targets/test_mysql_db/tasks/state_present_absent.yml
laurent-indermuehle: No sorry. I did that when the tests were failing on python 3.6.

Should I revert that?

Andersson007: @laurent-indermuehle 

> No sorry. I did that when the tests were failing on python 3.6.

I'm curious why they fail if they are not related to the changes. If it's safe, we can leave it as is.

Andersson007 (approved):

Andersson007:
@bmalynovytch thanks for reviewing!
I'll create an issue to change the default in the next major release.

PR #416, opened 2022-08-03
Fix ci python requirements

laurent-indermuehle:
##### SUMMARY

Lately, several CI tests are failing. It seems this is because Ansible 2.13 and devel, require Python >= 3.9 and we use 3.8 everywhere.

This pull request add  new test matrix to the Github Actions Plugin workflow and exclude all combinations that makes no sense (ansible 2.11 with Python 3.9 for instance).

PR #418, opened 2022-08-05
[PR #416/97318559 backport][stable-1] Fix ci python requirements

laurent-indermuehle:
* Add matrix for python and ansible-core versions for sanity tests

* Add python 3.9 to integrations tests

* Add python 3.9 to unit tests

* Reformat sort by python version first

(cherry picked from commit 97318559e5aa976dce668a62daaab258014d75a9)

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -29,6 +29,18 @@ jobs:
           - stable-2.12
           - stable-2.13
           - devel
+        python:
+          - 3.8
+          - 3.9
+        exclude:
+          - python: 3.8
+            ansible: stable-2.13
+          - python: 3.8
+            ansible: devel
+          - python: 3.9
+            ansible: stable-2.11
+          - python: 3.9
+            ansible: stable-2.12
     steps:
 
       - name: Check out code
@@ -39,7 +51,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: ${{ matrix.python }}
 
       - name: Install ansible-base (${{ matrix.ansible }})
         run: pip install https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz --disable-pip-version-check
@@ -64,6 +76,8 @@ jobs:
           - devel
         python:
           - 3.6
+          - 3.8
+          - 3.9
         connector:
           - pymysql==0.7.10
           - pymysql==0.9.3
@@ -71,14 +85,22 @@ jobs:
         exclude:
           - mysql: 8.0.22
             connector: pymysql==0.7.10
-          - python: 3.8
-            ansible: stable-2.11
           - python: 3.6
             ansible: stable-2.12
           - python: 3.6
             ansible: stable-2.13
           - python: 3.6
             ansible: devel
+          - python: 3.8
+            ansible: stable-2.11
+          - python: 3.8
+            ansible: stable-2.13
+          - python: 3.8
+            ansible: devel
+          - python: 3.9
+            ansible: stable-2.11
+          - python: 3.9
+            ansible: stable-2.12
     steps:
 
       - name: Check out code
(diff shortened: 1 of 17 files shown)


laurent-indermuehle:
After I've compared stable-1 and stable-2 I decided I can merge my conflict resolution from stable-2 to stable-1.

But now, the CI si failing with errors never seen before.

I've no clue what is going on. Maybe this errors were already there. I'll need to check older runs. Don't have time to do that until next week.

Also I'm wondering if it makes sense to test stable-1 against 2.13 and devel... Ho long is the the support for this branch?

Andersson007:
>Also I'm wondering if it makes sense to test stable-1 against 2.13 and devel... Ho long is the the support for this branch?

I think it makes sense as the collection (supported versions) have to work with all of the supported versions of ansible-core.
The rule is that we support the current version + 2 previous (i.e. in this case 3, 2, 1). When c.mysql 4.0.0 is released, we'll drop support for version 1.

Andersson007:
@laurent-indermuehle I think if we test against stable-2 and `main`, let's make it just working here with `stable-1` by limiting the scope excluding devel and stable-2.13. It should be fine as we test against all supported branches in the other branches. Feel free to exclude them.

laurent-indermuehle:
Thanks @Andersson007 for the proposed solution.
I'll try that soon.

laurent-indermuehle:
Excluding 2.13 and devel would not have solved the issue because the integrations' tests are failing also when using 2.12.
So I decided to give a try to fix the error:  1292, \"Truncated incorrect max_connect_errors value: '-1'\
Will see in a few hours if I'm correct.

laurent-indermuehle:
Silly me, it's a test intended to fail.
The issue is the assert changed == true that receive a false. But a failed task can return changed : true ?
I've checked, this happen only in stable-1
Just pushed 2 fixes, will see tomorrow.

laurent-indermuehle:
I checked the variable "changed" inside [tests/integration/targets/test_mysql_variables/tasks/assert_var.yml](https://github.com/ansible-collections/community.mysql/blob/stable-1/tests/integration/targets/test_mysql_variables/tasks/assert_var.yml) and it is "undefined".
Why?

Andersson007:
> I checked the variable "changed" inside [tests/integration/targets/test_mysql_variables/tasks/assert_var.yml](https://github.com/ansible-collections/community.mysql/blob/stable-1/tests/integration/targets/test_mysql_variables/tasks/assert_var.yml) and it is "undefined". Why?

good question:)

laurent-indermuehle:
To reproduce the issue, the file https://github.com/ansible-collections/community.mysql/blob/stable-1/tests/integration/targets/setup_mysql/vars/main.yml must be modified as such:

```yaml
# python_packages: [pymysql == 0.9.3]
python_packages: [mysqlclient == 2.0.1]
```

laurent-indermuehle:
Question, is older version of Ansible had a different assert module? I ask this, because the Jinja templating mustaches seams unnecessary to me here:

```yaml
assert:
    that:
      - "output.changed == {{ changed }}"
```
And in fact, if I remove them, I then get the following error on every tests :

```yaml
TASK [test_mysql_variables : assert output message changed value] ************************************************************************************************************************
fatal: [testhost]: FAILED! => {
    "assertion": "output.changed == changed",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

Andersson007:
if the error's scope is limited, you can just remove it. I don't think the error is related to our code

laurent-indermuehle:
Ok, Now i'm left with only pymsql 0.7 and 0.9 that fails to connect to the sandboxed database. login_user and login_password is rejected. Then it fails to find a /root/.my.cnf (which is normal because it is located in "{{ dbdeployer_sandbox_home_dir }}/multi_msb_{{ install_version|replace('.','_') }}/node1/my.sandbox.cnf" which translate to /root/sandboxes/multi_msb_8_0_22/node1/my.sandbox.cnf).

But why mysqlclient succeed all tests? That makes no sense :(

Andersson007:
@laurent-indermuehle thanks for fixing the tests!

laurent-indermuehle:
You're welcome. And I'm not done ;) 
https://github.com/laurent-indermuehle/community.mysql/commits/port-stable-1-ci-changes

PR #423, opened 2022-08-15
Port stable 1 ci changes

laurent-indermuehle:
This is a "frontport" of the modifications done directly in stable-1 while debugging the CI in #418.

My hope is to reduce merge conflict in the future by porting those modifications to main and stable-2.

PR #425, opened 2022-08-16
[PR #423/61586ae4 backport][stable-1] Port stable 1 ci changes (#423)

laurent-indermuehle:
This is merge conflict resolution from #416 copied from stable-1 to main and then backport in stable-1 again.

### DIFF
--- a/tests/integration/targets/test_mysql_variables/tasks/assert_var_output.yml
+++ b/tests/integration/targets/test_mysql_variables/tasks/assert_var_output.yml
@@ -35,6 +35,6 @@
 - name: assert output variable info match mysql variable info
   assert:
     that:
-      - "result.changed == true"
+      - "result.changed | bool == true"
       - "key_name in result.stdout"
       - "key_value in result.stdout"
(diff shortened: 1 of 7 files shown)


tests/integration/targets/test_mysql_variables/tasks/assert_var_output.yml
```diff
@@ -35,6 +35,6 @@
 - name: assert output variable info match mysql variable info
   assert:
     that:
-      - "result.changed == true"
+      - "result.changed | bool == true"
```
Andersson007, line 38:
```suggestion
      - "result.changed | bool == true"
```
i wonder if `result is changed` works here ? It must, i think, can we check?

laurent-indermuehle: Good point!

Mmh, I must recall how I made it fails in the first place. I think it was using Python 3.8 on stable-1 and target test_mysql_variables.

I let this PR open as a reminder to check that.

PR #442, opened 2022-09-09
add an option to use a specific service with pam plugin

hubiongithub:
##### SUMMARY

add an additional parameter for mysql_user to be able to use create user .. identified with ... USING ...

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

mysql_user

##### ADDITIONAL INFORMATION

```paste below
  mysql_user:
    name: username
    plugin: auth_pam
    plugin_auth_service_string: pam_mysql
...

now produces 
'CREATE USER 'username'@'localhost' IDENTIFIED WITH 'pam' USING 'pam_mysql' REQUIRE SSL',0

So pam exactly know which service is to be used.
This may also be an option for other auth_plugins, I just checked against pam here.
```

### DIFF
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -133,7 +133,7 @@ def get_existing_authentication(cursor, user):
 
 
 def user_add(cursor, user, host, host_all, password, encrypted,
-             plugin, plugin_hash_string, plugin_auth_string, new_priv,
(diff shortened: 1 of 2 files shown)


hubiongithub:
The way mariadb stores this information (using the authentication_string column) will need a lot more changes to the logic around handling updates ... of users specific to pam plugin.

PR #444, opened 2022-09-13
Fix: devel requires python 3.9 in roles CI

rsicart:
Fix [CI for roles](https://github.com/ansible-collections/community.mysql/actions/runs/3022728901/jobs/4862148734).

Package 'ansible-core' requires a different Python: 3.8.13 not in '>=3.9'

### DIFF
--- a/.github/workflows/ansible-test-roles.yml
+++ b/.github/workflows/ansible-test-roles.yml
@@ -32,8 +32,6 @@ jobs:
           - 3.8
           - 3.9
         exclude:
-          - python: 3.8
-            ansible: stable-2.11
           - python: 3.6
             ansible: stable-2.12
           - python: 3.6
@@ -41,7 +39,15 @@ jobs:
           - python: 3.6
             ansible: devel
           - python: 3.8
+            ansible: stable-2.11
+          - python: 3.8
+            ansible: stable-2.13
+          - python: 3.8
             ansible: devel
+          - python: 3.9
+            ansible: stable-2.11
+          - python: 3.9
+            ansible: stable-2.12
 
     steps:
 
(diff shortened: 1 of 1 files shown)


rsicart:
Hey guys,

@laurent-indermuehle @betanummeric @Jorge-Rodriguez @bmalynovytch 

someone wanna review those changes?

Thanks in advance!

laurent-indermuehle:
hehe, @rsicart  thanks for your motivation :)
I did the same for plugins CI here: https://github.com/ansible-collections/community.mysql/pull/416
As you can see, I had to exclude older version of Ansible when testing Python 3.9. So if I'm correct, you must do the same here.

I think it's difficult to know what works and what fails because this test is empty. I recall @Andersson007 explaining that the role CI was made available for the day we will host roles.

rsicart:
Hey @laurent-indermuehle, are you sure?

Actually all Roles CI molecule checks are successful 🤔

laurent-indermuehle:
@rsicart I think there are 2 things:

- Testing latest Python with an old Ansible doesn't make much sense and add time to the CI run, which is already way too long in my opinion.
- This is what this workflow tests: https://github.com/ansible-collections/community.mysql/blob/main/.github/workflows/ansible-test-roles.yml#L62 The answer is "nothing". In that case it easy to be green.

For the second part, that's ok. If someday we figure out a way to add roles, we will debug the CI then. But for the first part, I would prefer to keep this test to a minimum duration.

Andersson007 (approved):
@rsicart thanks for the contribution!
@laurent-indermuehle thanks for reviewing! Do you have any other ideas about the PR? If not, feel free to merge and backport

laurent-indermuehle:
@Andersson007 Ok, l'll do that. This PR align the roles tests with the ones for plugins. I should have done that in #416.

laurent-indermuehle:
This is the first backport that succeed on the first try. Congratz @rsicart :)

PR #473, opened 2022-12-09
CI: add PR change detection

Andersson007:
##### SUMMARY
CI: add PR change detection

Relates to:
- https://github.com/ansible-community/ansible-test-gh-action/pull/49
- https://github.com/ansible-community/ansible-test-gh-action#pull-request-change-detection

PR #474, opened 2022-12-09
[DO NOT MERGE] Test commit

Andersson007:
##### SUMMARY
Relates to https://github.com/ansible-collections/community.mysql/pull/473

### DIFF
--- a/plugins/modules/mysql_info.py
+++ b/plugins/modules/mysql_info.py
@@ -224,7 +224,6 @@ from ansible.module_utils._text import to_native
 # ===========================================
 # MySQL module specific support methods.
 #
-
 class MySQL_Info(object):
 
     """Class for collection MySQL instance information.


Andersson007:
@felixfontein I added `pull-request-change-detection: true` in https://github.com/ansible-collections/community.mysql/pull/473 and opened this test PR to check if now it runs only an affected target.
Seems all the targets run anyway.
Is it something unexpected or i misunderstood something?

felixfontein:
@Andersson007 if you check out the sanity and unit tests, you can see that it works fine.

For the integration tests it also works fine: the tests also modify some more files https://github.com/ansible-collections/community.mysql/blob/main/.github/workflows/ansible-test-plugins.yml#L18-L19, and the changes in these files trigger all tests. So ansible-test is doing exactly what it is supposed to do when doing change detection.

Andersson007:
> @Andersson007 if you check out the sanity and unit tests, you can see that it works fine.
> 
> For the integration tests it also works fine: the tests also modify some more files https://github.com/ansible-collections/community.mysql/blob/main/.github/workflows/ansible-test-plugins.yml#L18-L19, and the changes in these files trigger all tests. So ansible-test is doing exactly what it is supposed to do when doing change detection.

Ah, thanks for clarifying:)

PR #490, opened 2023-01-23
Integrations tests : Use containers for more control and verify that versions match expectation

laurent-indermuehle:
##### SUMMARY

Improve tests reliability:

- Test more versions of MariaDB
- Carefully verify every component of the tests in the new target 'setup_controller' to ensure expected versions are correct : Python, Ansible, connector and MySQL/MariaDB.
- Add tools to test locally the same as on GHA by using same containers and virtualenv. Custom test containers are published in ghcr.io by this repo's workflows. MySQL/MariaDB are official Docker Hub images.
- New name for many tasks to makes it easier to find failing tests: Rename duplicates. Add name for tasks which doesn't had one, refactor some tests files to better group tests by subject, ...

##### ISSUE TYPE

- Integration tests

##### COMPONENT NAME

CI / Integration tests

##### ADDITIONAL INFORMATION

The first goal was to add all latest major versions of MariaDB. Mainly the Long Term Support 10.6 which I thought wasn't working (spoiler, it was because I installed PyMySQL 1.0.2). While I was at it, I added 10.5 too. I also inserted 10.4 but we will remove it in late 2023 when EoL.

Secondly, I sometimes run into situations were the tests title lies about versions tested. I recall one time using tmate to SSH into a GHA Ubuntu VM to find out a MySQL 8 instead of the MariaDB 10.3 in the title of the test. 
I also lost so much time when working on this PR because I thought I was testing pymysql but was in fact running mysqlclient. So I say: **Never again!** and the file `verify.yml` in the target 'setup_controller' checks every component announced by the workflow. If the test say "I will test python 3.10", we go into the target container and do a "python3 -V" and check if the expected version is returned. In addition to Python, it also verify the version of Ansible, the connector and the database.

Finally, by using custom containers from ansible-test we have more control on what we install. I still use Ubuntu as base image and change the release to get the right mysqlclient and Python. For instance Ubuntu 18.04 provides mysql-client-5.7,mariadb-client-10.3 and Ubuntu 20.04 and 22.04 provides mysql-client-8,mariadb-client-10.6. Also availables Python vary between those versions. So this image include 3 things : 1) Python, 2) the Python connector and its dependencies and 3) mysql-client which provides the commands `mysql` and `mysqldump`. These custom containers have each a workflow to build and publish its images in ghcr.io. Those images can be found in the GitHub `<> Code` page of this repo in the "Packages" section.

[...]

PR #497, opened 2023-01-26
mysql_info - Add connector_name and connector_version to returned value

laurent-indermuehle:
##### SUMMARY

Add two returned values for the `mysql_info` plugin : `connector_name` and `connector_version`.
You may have both mysqlclient and pymysql installed. This helps you know which one was used.

This feature is mandatory for #490 who will add a step that verify that all components have expected versions at the start of integrations tests.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

mysql_info

##### ADDITIONAL INFORMATION

```paste below
# Before                                                                                                                                                                              
    "changed": false,                                                                                                                                                                     
    "databases": {                                                                                                                                                                        
        "information_schema": {                                                         
            "size": 212992                                          
        },                  

# After                                                                                                                                                                           
    "changed": false,
    "connector_name": "pymysql",
    "connector_version": "0.9.3",                                                                                                                                                                     
    "databases": {                                                                                                                                                                        
        "information_schema": {                                                         
            "size": 212992                                          
        },            
```

### DIFF
--- a/changelogs/fragments/497_mysql_info_returns_connector_name_and_version.yml
+++ b/changelogs/fragments/497_mysql_info_returns_connector_name_and_version.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - mysql_info - add connector_name and connector_version to returned values
--- a/changelogs/fragments/497_mysql_info_returns_connector_name_and_version.yml
+++ b/changelogs/fragments/497_mysql_info_returns_connector_name_and_version.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - mysql_info - add connector_name and connector_version to returned values
+  - mysql_info - add ``connector_name`` and ``connector_version`` to returned values (https://github.com/ansible-collections/community.mysql/pull/497).
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -67,7 +67,7 @@ def get_driver_version(mysql_driver):
     if driver_name == 'pymysql':
         # pymysql has two methods:
         # - __version__ that returns the string: 0.7.11.None
-        # - VERSION that returns the tupple (0, 7, 11, None)
+        # - VERSION that returns the tuple (0, 7, 11, None)
         v = mysql_driver.VERSION[:3]
         return '.'.join(map(str, v))
     elif driver_name == 'MySQLdb':
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -35,6 +35,8 @@ mysql_driver_fail_msg = ('A MySQL module is required: for Python 2.7 either PyMy
                          'Consider setting ansible_python_interpreter to use '
                          'the intended Python version.')
 
+from ansible_collections.community.mysql.plugins.module_utils.database import mysql_quote_identifier
+
 
 def get_driver_name(mysql_driver):
     """ (class) -> str
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -70,11 +70,12 @@ def get_driver_version(mysql_driver):
         # - VERSION that returns the tupple (0, 7, 11, None)
         v = mysql_driver.VERSION[:3]
         return '.'.join(map(str, v))
-
-    if driver_name == 'MySQLdb':
+    elif driver_name == 'MySQLdb':
         # version_info returns the tuple (2, 1, 1, 'final', 0)
         v = mysql_driver.version_info[:3]
         return '.'.join(map(str, v))
+    else:
+        return 'Unknown'
 
 
 def parse_from_mysql_config_file(cnf):
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -64,9 +64,6 @@ def get_driver_version(mysql_driver):
 
     driver_name = get_driver_name(mysql_driver)
 
-    if driver_name == 'Unknown':
-        return 'Unknown'
-
     if driver_name == 'pymysql':
         # pymysql has two methods:
         # - __version__ that returns the string: 0.7.11.None
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -38,41 +38,41 @@ mysql_driver_fail_msg = ('A MySQL module is required: for Python 2.7 either PyMy
 from ansible_collections.community.mysql.plugins.module_utils.database import mysql_quote_identifier
 
 
-def get_driver_name(mysql_driver):
+def get_connector_name(connector):
     """ (class) -> str
-    Return the name of the driver (pymysql or mysqlclient (MySQLdb))
-    or 'Unknown' if the driver name is not pymysql or MySQLdb. When adding a
-    connector here, also modify get_driver_version.
+    Return the name of the connector (pymysql or mysqlclient (MySQLdb))
+    or 'Unknown' if not pymysql or MySQLdb. When adding a
+    connector here, also modify get_connector_version.
     """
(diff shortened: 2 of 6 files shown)


changelogs/fragments/497_mysql_info_returns_connector_name_and_version.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - mysql_info - add connector_name and connector_version to returned values
```
Andersson007:
```suggestion
  - mysql_info - add ``connector_name`` and ``connector_version`` to returned values (https://github.com/ansible-collections/community.mysql/pull/497).
```

plugins/module_utils/mysql.py
```diff
@@ -37,6 +38,48 @@
...
+    if mysql_driver.__name__ not in ['pymysql', 'MySQLdb']:
+        return 'Unknown'
+
+    return mysql_driver.__name__
+
+
+def get_driver_version(mysql_driver):
+    """ (class) -> str
+    Return the version of pymysql or mysqlclient (MySQLdb).
+    If the driver name is unknown, this method also return 'Unknown'
```
Andersson007:
```suggestion
    If the driver name is unknown, this method also return 'Unknown'.
```

plugins/modules/mysql_info.py
```diff
@@ -206,6 +206,19 @@
   type: dict
   sample:
   - { "2": { "Host": "", "Master_id": 1, "Port": 3306 } }
+connector_name:
+  description: Name of the python connector used by the plugin. When the driver is not identified, returns C(Unknown).
```
```suggestion
  description: Name of the python connector used by the module. When the connector is not identified, returns C(Unknown).
```
i suggest using one of the term. If you prefer `driver` instead of `connector`, let's use it. If connector is fine, could you please change the function names to and other occurances?

plugins/modules/mysql_info.py
```diff
@@ -206,6 +206,19 @@
...
+  returned: always
+  type: str
+  sample:
+  - "pymysql"
+  - "MySQLdb"
+connector_version:
+  description: Version of the python connector used by the plugin. When the driver is not identified, returns C(Unknown).
```
```suggestion
  description: Version of the python connector used by the module. When the connector is not identified, returns C(Unknown).
```

plugins/modules/mysql_info.py
```diff
@@ -206,6 +206,19 @@
   type: dict
   sample:
   - { "2": { "Host": "", "Master_id": 1, "Port": 3306 } }
+connector_name:
+  description: Name of the python connector used by the plugin. When the driver is not identified, returns C(Unknown).
+  returned: always
+  type: str
+  sample:
+  - "pymysql"
+  - "MySQLdb"
```
Andersson007, line 216:
```suggestion
  - "MySQLdb"
  version_added: '3.6.0'
```

plugins/modules/mysql_info.py
```diff
@@ -206,6 +206,19 @@
...
+connector_version:
+  description: Version of the python connector used by the plugin. When the driver is not identified, returns C(Unknown).
+  returned: always
+  type: str
+  sample:
+  - "1.0.2"
```
Andersson007, line 223:
```suggestion
  - "1.0.2"
  version_added: '3.6.0'
```

plugins/module_utils/mysql.py
```diff
@@ -37,6 +38,48 @@
...
+    """
+
+    if mysql_driver is None:
+        return 'Unknown'
+
+    driver_name = get_driver_name(mysql_driver)
+
+    if driver_name == 'Unknown':
+        return 'Unknown'
+
```
Andersson007:
```suggestion
```

plugins/module_utils/mysql.py
```diff
@@ -37,6 +38,48 @@
...
+        # pymysql has two methods:
+        # - __version__ that returns the string: 0.7.11.None
+        # - VERSION that returns the tupple (0, 7, 11, None)
+        v = mysql_driver.VERSION[:3]
+        return '.'.join(map(str, v))
+
+    if driver_name == 'MySQLdb':
+        # version_info returns the tuple (2, 1, 1, 'final', 0)
+        v = mysql_driver.version_info[:3]
+        return '.'.join(map(str, v))
```
Andersson007:
```suggestion
    elif driver_name == 'MySQLdb':
        # version_info returns the tuple (2, 1, 1, 'final', 0)
        v = mysql_driver.version_info[:3]
        return '.'.join(map(str, v))
    else:
        return 'Unknown'
```

Andersson007 (commented):
@laurent-indermuehle great job, thank you!

Could you please add integration tests?
If it's impossible or hard to determine what the new values return in each particular case we could check if they return anything from the possible options using `or`. For versions it can be just a check that it's not none.

plugins/modules/mysql_info.py
laurent-indermuehle: You mean refactor all the code to change "driver" to "connector" everywhere? So even `mysql_driver_fail_msg` -> `mysql_connector_fail_msg`? It's only 99 occurrences, but I want to be sure I understood correctly before.

Andersson007: i think only the code you added would be fine. but you can use `driver` instead of `connector`, up to you. the goal is to make it consistent in terms of terminology. any of them is fine

Andersson007: i would leave legacy code as is

plugins/module_utils/mysql.py
```diff
@@ -37,6 +38,46 @@
...
+    """ (class) -> str
+    Return the name of the connector (pymysql or mysqlclient (MySQLdb))
+    or 'Unknown' if not pymysql or MySQLdb. When adding a
+    connector here, also modify get_connector_version.
+    """
+    if connector is None or not hasattr(connector, '__name__'):
+        return 'Unknown'
+
+    if connector.__name__ not in ['pymysql', 'MySQLdb']:
+        return 'Unknown'
```
Andersson007:
```suggestion
    if connector.__name__ not in ['pymysql', 'MySQLdb']:
        return 'Unknown'
```
should we maybe return anything it returns? i.e. remove this statement?

Andersson007: e.g. if in future they, say, decide to return `mysqlclient` or whatever it'll be `Unknown`

laurent-indermuehle:
Green! At last! :)
Thanks for the suggestions @Andersson007, I think I've done all of them.

Andersson007 (approved):
@laurent-indermuehle great addition! Thanks!

How about adding yourself to `author:` ? (feel free to merge once you add yourself or now)

laurent-indermuehle:
> @laurent-indermuehle great addition! Thanks!
> 
> How about adding yourself to `author:` ? (feel free to merge once you add yourself or now)

Where should I do that? In the documentation block of `plugins/modules/mysql_info.py`?

Andersson007:
> > @laurent-indermuehle great addition! Thanks!
> > How about adding yourself to `author:` ? (feel free to merge once you add yourself or now)
> 
> Where should I do that? In the documentation block of `plugins/modules/mysql_info.py`?

@laurent-indermuehle yep

laurent-indermuehle:
Ok thanks.

@Andersson007 Now for the difficult question. Should I backport this new feature?
I think yes because it's a requirements for #490 that I also want to backport. Otherwise we will have 2 differents testings workflows.

Andersson007:
> Ok thanks.
> 
> @Andersson007 Now for the difficult question. Should I backport this new feature? I think yes because it's a requirements for #490 that I also want to backport. Otherwise we will have 2 differents testings workflows.

@laurent-indermuehle as it's an exceptional case i think we can backport it to stable-2

PR #522, opened 2023-03-30
Add filter to prevent rebuild containers on push on stable-(1|2)

laurent-indermuehle:
##### SUMMARY

When we backport changes to stable-1 and stable-2, the workflows responsible to build the containers images are triggered. This is useless because if we backport from main, it means the containers are already published! No need to overwrite them (almost 2GB each...)

Thus I propose to filter out `stable-*` branches from the triggers.

##### ISSUE TYPE

- CI Pull Request

##### COMPONENT NAME

GHA

##### ADDITIONAL INFORMATION

```yaml
on:
  push:
    paths:
      - 'test-containers/mysql-py39-pymysql093/*'
      - '.github/workflows/docker-image-mysql-py39-pymysql093.yml'
      - '.github/workflows/build-docker-image.yml'
    branches-ignore:
      - stable-*
```
I think adding `branches-ignore` in the `push` trigger is enough. 

Or maybe it's best to trigger only on `main`? Then if someone changes the images on his fork, the images won't be rebuilt while opening a PR.

But how to update the containers? Do you need to remove the filter on the 15 workflows? And remember to put them back afterward?

Do anyone have experiences with triggers on commit messages? Like this:

```yaml
jobs:
  call-workflow-passing-data:
    if: "contains(github.event.head_commit.message, '[publish]')"
```

### DIFF
--- a/.github/workflows/docker-image-mariadb-py310-mysqlclient211.yml
+++ b/.github/workflows/docker-image-mariadb-py310-mysqlclient211.yml
@@ -7,6 +7,8 @@ on:
       - 'test-containers/mariadb-py310-mysqlclient211/**'
       - '.github/workflows/docker-image-mariadb-py310-mysqlclient211.yml'
       - '.github/workflows/build-docker-image.yml'
+    branches-ignore:
+      - stable-*
 
 jobs:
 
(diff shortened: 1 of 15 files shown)


Andersson007:
i unfortunately have no experience with this and resp idea. trust your judgment

laurent-indermuehle:
The best way to know is by experience. So next time the containers needs to be changed, we can adapt the triggers.
For now, I'll merge and backport because I'm sure the `branches-ignore: - stable-*` is useful.

PR #533, opened 2023-04-25
feat: Add support for the connector pymysql 1.0.2

laurent-indermuehle:
##### SUMMARY

This PR add the connector PyMySQL 1.0.2 to the integration tests matrix.

I know PyMySQL 1.0.3 is here since 2 months, but the test containers for 1.0.2 was already present so I use that version.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

All modules

##### ADDITIONAL INFORMATION

I had to disable tests for "changed" state in two places. This is not ideal but the only solution would be to patch the connectors to allow to read the warning from a query with `IF EXISTS`. But I took care to document that PyMySQL stopped returning a correct `changed` status since version 0.10.0.

Unless we run a new query `SHOW WARNINGS;`  after each query? If someone think it's possible, I'll happy to try this in another PR.

```paste below

```

### DIFF
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -139,6 +139,9 @@ jobs:
             python: '3.9'
 
           - python: '3.8'
+            connector_version: 1.0.2
+
+          - python: '3.8'
             connector_version: 2.0.3
 
           - python: '3.8'
@@ -146,6 +149,9 @@ jobs:
 
           - python: '3.9'
             connector_version: 0.7.11
+
+          - python: '3.9'
+            connector_version: 1.0.2
 
           - python: '3.9'
             connector_version: 2.0.1
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -64,15 +64,7 @@ jobs:
         connector_version:
           - 0.7.11
           - 0.9.3
-          # Before we can activate test with pymysql 1.0.2 we should debug the
-          # following plugins:
-          #
-          # mysql_query:
-          # test "Assert that create table IF NOT EXISTS is not changed with pymysql" failed
-          #
-          # mysql_replication:
-          # test "Assert that startreplica is not changed" failed
-          # - 1.0.2
+          - 1.0.2
           - 2.0.1
           - 2.0.3
           - 2.1.1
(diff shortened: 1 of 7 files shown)


Andersson007 (approved):
@laurent-indermuehle LGTM thansk a lot!

1. Trying to use `SHOW WARNINGS;` SGTM
2. This though just came to mind: do we really need to test against 0.Y.Z versions? They are dev versions, all up to 1.Y.Z. Ideas?

laurent-indermuehle:
@Andersson007 :
- I'm adding a issue for the `SHOW WARNING;` idea.
- RHEL8 and RHEL9 both ship PyMySQL 0.10.1 in their RPM repositories. Preferably we should test this version instead of 0.7, 0.9 and 1.0. Or hunt down the person responsible for this RPM and beg him/her to release 1.0.x ;)

Andersson007:
@laurent-indermuehle i wouldn't orient much on what RHEL ships by default:)
I'm a bit lost these days, so my comments can be far from reality and sound like nonsense due to sickness:) Like the following one:
Could we remove 0.9.3 from the matrix in this PR to reduce CI load a bit?
Generally, I think testing against 0.7 is definitely enough (or against the oldest supported version which i don't know) :)

laurent-indermuehle:
@Andersson007 I don't know the history of the collection. I only know what I saw on my RHEL server when I install the connectors available to me.

Since we test every major release of:
- Ansible
- Mysql
- Mariadb
- Python 
I think it would be fair to also tests every latest major release of our connectors:

PyMySQL https://pypi.org/project/pymysql/#history:
- 0.10.1 <- This happens to be the version I'm interested in the most :)
- 1.0.3

Mysqlclient https://pypi.org/project/mysqlclient/#history:
- 1.4.6 <- branch 1.x has never been tested AFAIK
- 2.1.1

In order to minimize the amount of test containers, I used a different version of each connectors per version of Python: 
- python 3.8 mysqlclient 2.0.1 pymysql0711 pymysql093
- python 3.9 mysqlclient 2.0.3 pymysql093
- python 3.10 mysqlclient 2.1.1 pymysql102
So we are not testing every Python with every connectors.

Also, we are soon going to add:
- Mariadb 10.11
- Mariadb 11.1
- Python 3.11
- Ansible 2.15

Shouldn't we transform this conversation into a GitHub discussion and merge this PR while we wait on the input of other users of the collection?

Andersson007:
@laurent-indermuehle 

>Shouldn't we transform this conversation into a GitHub discussion and merge this PR while we wait on the input of other users of the collection?

Yeah, let's do it

laurent-indermuehle:
I did: https://github.com/ansible-collections/community.mysql/discussions/541
I'm now merging this PR.

Andersson007:

PR #542, opened 2023-05-04
Release 2.4.1 commit

laurent-indermuehle:
Release 2.4.1 commit


Andersson007 (approved):
@laurent-indermuehle if you don't mind you could remove those Minor changes fragments and entries (as they are not minor changes actually). If you don't want doing releasing again, we could merge it as is.

Issue #546, opened 2023-05-08
Possible recurrence of issue #151 following latest release of community.mysql?


paulh-rtr:
##### SUMMARY
A Github Action which we run on a weekly basis, and which calls an Ansible playbook which in turn calls community.mysql, is failing since this weekend on all calls to community.mysql 

Nothing has changed in the codebase; this was confirmed as working last weekend, and so I presume the issue is with the most recent release 3.7.0.  Running the playbook by hand, in a different environment with an earlier version of community.mysql, works.

For example see below (with some details redacted):
```
- name: create heartbeat user
  community.mysql.mysql_user:
<snip>
    state: present
  when: db_role == "{{ master_role }}"
```

This produces an error like the following:
```
fatal: [<hostname>]: FAILED! => {"changed": false, "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: 'database' is an invalid keyword argument for this function"}
```

**Note:** The error message would seem to be the same as in this issue: https://github.com/ansible-collections/community.mysql/issues/151.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
mysql_user, mysql_query, possibly others

##### ANSIBLE VERSION
```
ansible.posix:1.5.2
```
##### COLLECTION VERSION
```
community.mysql:3.7.0
```

##### OS / ENVIRONMENT
Github Action environment as described below:
```
Current runner version: '2.304.0'
Runner name: 'infra-runners-default-tr8[2](https://github.com/RentTheRunway/mysql/actions/runs/4904303710/jobs/8757254064#step:1:2)k-cqkkn'
Runner group name: 'infra-runners'
Machine name: 'infra-runners-default-tr82k-cqkkn'
GITHUB_TOKEN Permissions

Starting galaxy role install process
- downloading role 'google_cloud_ops_agents', owned by googlecloudplatform
- downloading role from https://github.com/GoogleCloudPlatform/stackdriver-ansible-role/archive/master.tar.gz
- extracting googlecloudplatform.google_cloud_ops_agents to /home/runner/.ansible/roles/googlecloudplatform.google_cloud_ops_agents
- googlecloudplatform.google_cloud_ops_agents (master) was installed successfully
Starting galaxy collection install process
Process install dependency map
Starting collection install process
[...]

PR #551, opened 2023-05-16
fix connection arguments mysql driver compatability

betanummeric:
bugfix for #546, relates to #116

### what arguments are supported in what drivers

- pymysql has deprecated the `db` argument in [1.0.0](https://github.com/PyMySQL/PyMySQL/blob/main/CHANGELOG.md#v100) (2021), but it will probably still work for some time. I'm not sure since when pymysql supports the `database` argument, but at least since 0.7.11, which we test.
- mysql-python (MySQLdb) only seems to support `db`. The project was last changed in 2014 so we could discuss if we want to support it. The current issue is about this old driver, see https://github.com/ansible-collections/community.mysql/issues/546#issuecomment-1544078481.
- The successor of mysql-python, mysqlclient, also deprecated the `db` argument in [2.1.0](https://github.com/PyMySQL/mysqlclient/releases/tag/v2.1.0) (2021), like pymysql.

The same way, `passwd` is deprecated in favor of `password`.

### proposed fix

Check the mysql driver version to determine what argument to use. Use `db` and `passwd` until it's deprecated, then `database` and `password`.

### DIFF
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -144,18 +144,24 @@ def mysql_connect(module, login_user=None, login_password=None, config_file='', 
     if _mysql_cursor_param == 'cursor':
         # In case of PyMySQL driver:
         if mysql_driver.version_info[0] < 1:
-            # for PyMySQL < 1.0.0, use 'db' instead of 'database'
+            # for PyMySQL < 1.0.0, use 'db' instead of 'database' and 'passwd' instead of 'password'
             if 'database' in config:
                 config['db'] = config['database']
                 del config['database']
+            if 'password' in config:
+                config['passwd'] = config['password']
+                del config['password']
         db_connection = mysql_driver.connect(autocommit=autocommit, **config)
     else:
         # In case of MySQLdb driver
         if mysql_driver.version_info[0] < 2 and mysql_driver.version_info[1] < 1:
-            # for MySQLdb < 2.1.0, use 'db' instead of 'database'
+            # for MySQLdb < 2.1.0, use 'db' instead of 'database' and 'passwd' instead of 'password'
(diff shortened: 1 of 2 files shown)


plugins/module_utils/mysql.py
```diff
@@ -143,9 +143,25 @@ def mysql_connect(module, login_user=None, login_password=None, config_file='',
...
+            if 'database' in config:
+                config['db'] = config['database']
+                del config['database']
+            if 'password' in config:
+                config['passwd'] = config['password']
+                del config['password']
         db_connection = mysql_driver.connect(autocommit=autocommit, **config)
     else:
         # In case of MySQLdb driver
+        if mysql_driver.version_info[0] < 2 and mysql_driver.version_info[1] < 1:
```
Andersson007, line 157:
```suggestion
        if mysql_driver.version_info[0] < 2 and mysql_driver.version_info[1] < 1:
```
should we use `get_connector_name` and `get_connector_version` + `LooseVersion` in the conditions? would it be more readable and safe?

Andersson007 (commented):
@betanummeric thanks for the PR!

plugins/module_utils/mysql.py
betanummeric: Using `get_connector_version` and `LooseVersion` would take the same data (`.version_info`), convert it to a string, parse it and compare it. I prefer the direct integer comparison as implemented in this PR. There is also a comment which should explain it.

`get_connector_name` is more intuitive than `_mysql_cursor_param`, I will use it.

Andersson007: Sure thing, thanks:)

Andersson007 (approved):
@betanummeric looks great, thank you!

PR #570, opened 2023-09-07
Mysql_db dump/import add missing SSL mode

FlorianPerrot:
##### SUMMARY
Add missing argument `--ssl` for db dump and import if `check_hostname` is true.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- plugins/modules/mysql_db.py

##### ADDITIONAL INFORMATION
Useful, if you want is ssl mode without certifcat.
https://mariadb.com/kb/en/securing-connections-for-client-and-server/#enabling-one-way-tls-for-mariadb-clients-without-server-certificate-verification

### DIFF
--- a/changelogs/fragments/add_ssl_mode_for_dump_and_import.yml
+++ b/changelogs/fragments/add_ssl_mode_for_dump_and_import.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - mysql_db - Implement `check_hostname` parameter for state dump and import.
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -371,7 +371,7 @@ def db_delete(cursor, db):
 
 
 def db_dump(module, host, user, password, db_name, target, all_databases, port,
-            config_file, socket=None, ssl_cert=None, ssl_key=None, ssl_ca=None,
+            config_file, socket=None, check_hostname=False, ssl_cert=None, ssl_key=None, ssl_ca=None,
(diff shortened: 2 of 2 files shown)


Andersson007:
@FlorianPerrot hello, thanks for the patch!
could you please add a changelog fragment https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment ?

betanummeric (changes requested):
The `--ssl` option is implied by other `--ssl-*` options. As a workaround, you could set the `ssl_ca` variable to some file with trusted CA certs.

The `--ssl` argument does not exist in MySQL 8, it was removed in favor of `--ssl-mode`.

The name `check_hostname` is ambiguous: What is checked?
I would expect it to mean `--ssl-verify-server-cert` for MariaDB and `--ssl-mode=VERIFY_IDENTITY` for MySQL. But this should be decided and documented.

MariaDB docs: https://mariadb.com/kb/en/securing-connections-for-client-and-server/#enabling-one-way-tls-for-mariadb-clients
MySQL docs: https://dev.mysql.com/doc/refman/8.0/en/using-encrypted-connections.html#using-encrypted-connections-client-side-configuration

FlorianPerrot:
Hum good point. I hadn't seen "--ssl" argument different with mariadb and mysql...
But I have a use case where it is useful to require "--ssl" for One-Way TLS without server certificate verification.
I think my feature is not necessary, I can use "config_file" parameters and "check_hostname"  is indeed ambiguous.

PR #569, opened 2023-09-06
Fix column uppercasing

kmarse:
##### SUMMARY
This pull request is to add logic to remove uppercasing column level privileges as they are case sensitive per MySQL databases.

This fixes #399 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

plugins/modules/mysql_role.py
plugins/modules/mysql_user.py
plugins/module_utils/user.py

##### ADDITIONAL INFORMATION

```paste below

```

Issue #399, opened 2022-06-13
Column Level Privileges are capitalized when applied to DB instance


kmarse:
Please raise issues via the [new interface](https://github.com/ansible/ansible/issues/new/choose)
Currently I use the mysql-roles module to manage all app roles in our db instances. I have some column level privileges to a larger table that I don't want to grant access to the whole table to some roles. I have been receiving access denied errors now because when the mysql-roles module applied the grants it capitalized the column names.

In MySQL's documentation I have found that it states that column level privileges are not case sensitive but after speaking to an oracle support rep he has confirmed that they are.

I am confused on why the module when inputting the grants for column level privileges is uppercasing the column name and not leaving it the way I have it in my playbook example follows:

What I have in my playbook:
'database.table': 'SELECT(`app_type`, `rec_id`)'

What MySQL is returning after application:
GRANT SELECT (`APP_TYPE`, `REC_ID`) ON `database`.`table` TO `role`@`%`

Andersson007:
@kmarse hello, thanks for reporting this!
I've taken a quick look at the code and i see that the issue is deeper and broader than just uppercasing the input as it relates to the way the collection handles priv-related data fetched from a DB- it just uppercases everything. The module is relatively new but it uses the old mysql_user's code.
I'll try to make time to find a solution but if someone wants to pick it up, please go ahead.

kmarse:
@Andersson007 Thanks for getting back to me on this. 

I did some investigating into the different modules and I don't see anywhere currently that it would be uppercasing the priv variable. Is there a place of reference you could point me to so that I can test a solution?

kmarse:
@Andersson007 I ran some tests and found the issue which I have resolved locally. 

The issue was found in user.py on line 644 which performs an upper() function on the entire privilege statement including column_names. 

output[pieces[0]] = re.split(r',\s*(?=[^)]*(?:\(|$))', pieces[1].upper())

This causes 'SELECT(`app_type`, `rec_id`)' to become 'SELECT(`APP_TYPE`, `REC_ID`)'. Privileges are reserved words in MySQL which means that SELECT will always be capitalized. I do understand that if some one in their privs variables are not passing uppercase privileges then ansible would report a change in grants every single time because the module would see GRANT select (`app_type`, `rec_id`) ON db.table TO user@host; but mysql databases would return when queried GRANT SELECT (`app_type`, `rec_id`) ON db.table TO user@host;

[...]

PR #572, opened 2023-09-15
feat[mysql_info]: add 'users_privs' filter

laurent-indermuehle:
##### SUMMARY

Add filter `users_privs` to the `mysql_info` module. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

`mysql_info`

##### ADDITIONAL INFORMATION

This filter returns information about users accounts. The output can be used as an input of the mysql_user plugin. Useful when migrating accounts to a new server or to create an inventory.

It doesn't return informations about `password_option` and `lock_option` but since the `mysql_user`module doesn't support them, I don't bother implemented them yet.

Doesn't support `sha256_password` and `caching_sha2_password` authentication plugins. Even if you use print_identified_with_as_hex to convert the password into a hash. I think this is because PyMySQL and
mysqlclient wrap the password in quote: `CREATE USER... IDENTIFIED AS '0x1234'` instead of `CREATE USER... IDENTIFIED AS 0x1234`. We could try to convert to another python connector, like mysql-connector-python. But this is way too much work for me who only work on MariaDB with `mysql_native_password`.

```paste below

```


laurent-indermuehle:
Ready for review please, if someone has the time, I would really appreciate another pair of eyes.
The only test that fails is sanity for devel which is the brand new ansible-core 2.17. We can fix the syntax for devel in a later PR.

changelogs/fragments/lie_mysql_info_users_privs.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+minor_changes:
+
+  - mysql_info - add filter users_privs
```
markuman:
```suggestion
  - mysql_info - add filter users_privs (https://github.com/ansible-collections/community.mysql/pull/572).
```

laurent-indermuehle:
I'm starting to use this feature (from my fork until next release) and noticed two errors in the keys I used for the output that didn't match what mysql_user is expecting as an input. It's now fixed.

I can now say it works perfectly. I just exported 300 accounts and re-create them using mysql_user in check mode. All green :)

plugins/modules/mysql_info.py
```diff
@@ -219,7 +220,8 @@
   description:
     Information about users accounts. The output can be used as an input of the
     mysql_user plugin. Useful when migrating accounts to a new server or to
-    create an inventory. Does not support proxy privileges.
+    create an inventory. Does not support proxy privileges. Cause issue with
+    authentications plugins sha256_password and caching_sha2_password.
```
```suggestion
    authentications plugins C(sha256_password) and C(caching_sha2_password).
```
to highlight them a bit

changelogs/fragments/lie_mysql_info_users_privs.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+minor_changes:
+
+  - mysql_info - add filter users_privs (https://github.com/ansible-collections/community.mysql/pull/572).
```
Andersson007:
```suggestion
  - mysql_info - add filter ``users_privs`` (https://github.com/ansible-collections/community.mysql/pull/572).
```

plugins/module_utils/mysql.py
```diff
@@ -194,6 +193,16 @@ def mysql_common_argument_spec():
...
 
 
+def get_server_type(cursor):
+    """ Return MySQL or MariaDB after quering the server
+    using SELECT VERSION()"""
+    srv_ver = get_server_version(cursor)
+    if 'mariadb' in srv_ver.lower():
+        return "mariadb"
+    else:
+        return "mysql"
```
Andersson007:
```suggestion
    return 'mariadb' if 'mariadb' in srv_ver.lower() else "mysql"
```
or sth like that

plugins/module_utils/user.py
```diff
@@ -477,7 +486,37 @@ def pick(x):
         else:
             return x
 
+    mysql8_all_privileges = [
```
```suggestion
    mysql8_all_privileges = [
```
if we can name these groups with meaningful names, wouldn't a dict serve better here? especially later when invoking the sublists? Also the second list look sorted

Andersson007 (commented):
@laurent-indermuehle great undertaking, thanks!

laurent-indermuehle:
Thanks @Andersson007 for your review.

Regarding your comment about MySQL 8 being already sorted, I copied the output into my editor. So MySQL must have sorted the privileges for me.

But I just discovered an issue with my implementation. Because I sorted the privileges, I get perma-yellow because the order of the privileges don't match the server output anymore.

What should I do? Fix my function to not sort the output, or fix mysql_user.changed to always sort privileges before comparison?

plugins/module_utils/user.py
laurent-indermuehle: @Andersson007 honestly I don't know. I barely understand the output of `show grants for root@localhost` on MySQL 8 (I only use MariaDB and MySQL 5.7.). Why 2 lines ? How to call the keys in the dictionary? "Basic" and "Advanced"?

laurent-indermuehle:
I've pushed a new test that fails due to uppercase fields as described in #399.
This PR is blocked and needs to be rebases onto #569. I'll try to work on that soon.

plugins/module_utils/user.py
Andersson007: @laurent-indermuehle call them as you like, though i would make them fully lowercased

Andersson007:
>What should I do? Fix my function to not sort the output, or fix mysql_user.changed to always sort privileges before comparison?

Sorting sounds like a safe solution to me

Andersson007:
@laurent-indermuehle i would also suggest implementing testing in a separate PR and then rebasing this one after we merge the testing-related one (i.e. not to keep not tightly related things together)

laurent-indermuehle:
Not sure why sanity 2.17 complains here, on my computer it works.

After a second look, I now see that I didn't sort the privileges! I do it only during comparison with `mysql8_all_privilege`. I've tested on 33 MariaDB servers, all reports all green! :)

I've changed `mysql8_all_privilege` to a dict as you proposed @Andersson007. I hope it's what you meant.

What do you mean by "testing in a separate PR" @Andersson007?

Andersson007:
@laurent-indermuehle 

>What do you mean by "testing in a separate PR" @Andersson007?

I think i saw some test matrix changes here earlier but now i don't:) so just ignore this statement

laurent-indermuehle:
@Andersson007 ah right, that was the case before I revert my commit were I merge main into that branch.

But now MySQL 8.0.31 fails and I don't know why :(

laurent-indermuehle:
I found what is the weird list of "ALL PRIVILEGES" MySQL 8 is spitting: They are either static (SELECT, INSERT, ...) or dynamic (APPLICATION_PASSWORD_ADMIN, ...). The dynamic privileges are only loaded if the plugin that use them is enabled. More information here: https://dev.mysql.com/doc/refman/8.0/en/privileges-provided.html#static-dynamic-privileges

So that explains why "SHOW GRANTS FOR" was displaying 2 lists.

The issue is that I broke everything for 8.0 : When you create a user with all privileges, it got created with only the dynamic privileges. All statics are overwritten by USAGE:

```SQL
GRANT USAGE ON *.* TO `user1`@`%`
GRANT APPLICATION_PASSWORD_ADMIN,[...] ON *.* TO `user1`@`%`
-- Instead of
GRANT SELECT, INSERT, UPDATE, DELETE, CREATE, [...] ON *.* TO `user1`@`%`
GRANT APPLICATION_PASSWORD_ADMIN,[...] ON *.* TO `user1`@`%`
```

We have 2 functions in `module_utils/mysql.py`: `user_add` and `user_mod`. I've a hard time figuring out why creating a brand new user in mysql_db integrations tests call the `user_mod` one. Here is the output of the tests:

```
TASK [test_mysql_db : State Present Absent | Create user1 to access database dbuser1] ***
changed: [testhost] => {"changed": true, "msg":  
"Privileges updated: granted [], revoked ['SHOW VIEW', 'INDEX', 'UPDATE', 'EVENT',   
'LOCK TABLES', 'CREATE TABLESPACE', 'CREATE USER', 'RELOAD', 'REPLICATION SLAVE',  
 'CREATE ROUTINE', 'SHOW DATABASES', 'CREATE ROLE', 'CREATE TEMPORARY TABLES',  
 'DELETE', 'ALTER', 'TRIGGER', 'SUPER', 'CREATE', 'ALTER ROUTINE', 'REFERENCES', 'INSERT',  
 'EXECUTE', 'FILE', 'SHUTDOWN', 'DROP ROLE', 'PROCESS', 'SELECT', 'DROP',  
 'REPLICATION CLIENT', 'CREATE VIEW']", "password_changed": false, "user": "user1"}
``` 
This is so wrong...

laurent-indermuehle:
@Andersson007 no need to search anymore. I just found the issue. Thanks anyway.

The tests use a loop. So the second iteration uses user_mod().
I'm adding an option to privileges_get() to not summarize ALL privileges when we need to compare actual privileges with the ones provided by Ansible.

And also, I used to disable many tests to speed up my feedback loop, but this time it masked many regressions I introduced. Slow tests makes me do stupid things -_-'

laurent-indermuehle:
This PR started before others PR, merged since, required for this one.
Also, I missed an important fact right at the beginning that made things very hard. I'll explain bellow.
So I'm closing this PR and opening a new one to start fresh with fewer commits and a minimum of changes.

So, for the explanation: The issue was that `mysql_user` instanciate a cursor with the class `Cursor`, while `mysql_info` chose the class `DictCursor`.
So anytime I was calling a method that do a `cursor.fetchone()`, I was getting a dictionary instead of a list. This explains the "KeyError 0" I was getting from almost every functions from `module_utils.mysql` and `module_utils.user`!

PR #579, opened 2023-10-11
cut option maria_role as it has nothing to do with roles

laurent-indermuehle:
##### SUMMARY

This option was introduced in #189. To my knowledge, there is no difference between MySQL and MariaDB regarding roles or when you call a user by its name alone. Same for roles.

For instance:
```SQL
create user foo;
select user, host from mysql.user;
-- foo, %
show grants for foo;
-- GRANT USAGE ON *.* TO foo@%;

create role role_foo;
select user, host from mysql.user;
-- foo, %
-- role_foo, %
show grants for role_foo;
```
So both in MySQL and MariaDB you can omit the host.

The `mysql_user` plugin defaults host to `localhost`: `host=dict(type='str', default='localhost')`
So all functions will always receive a host. Maybe the option was used to bypass the default, 'localhost' and search for '%' instead of localhost ? This works because when you don't specify a host, MySQL and MariaDB will assume host is '%'.
But then, I don't get why the search for '%' is made by an option called `maria_role`.

I removed this option and if all tests path, I propose the get rid of it.

Sometime we must know if we are in MySQL or MariaDB, like when parsing the return of `SHOW GRANTS FOR`. For those cases I added a new function to the module_utils `mysql.py` called `get_server_type()`.

##### ISSUE TYPE

- Code clarity Pull Request

##### COMPONENT NAME

- mysql_user
- mysql_role
- module_utils: mysql

### DIFF
--- a/plugins/module_utils/mysql.py
+++ b/plugins/module_utils/mysql.py
@@ -194,6 +194,13 @@ def mysql_common_argument_spec():
     )
 
 
+def get_server_type(cursor):
+    """ Return MySQL or MariaDB after quering the server
+    using SELECT VERSION()"""
+    srv_ver = get_server_version(cursor)
+    return 'mariadb' if 'mariadb' in srv_ver.lower() else "mysql"
+
+
 def get_server_version(cursor):
     """Returns a string representation of the server version."""
     cursor.execute("SELECT VERSION() AS version")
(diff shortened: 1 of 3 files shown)


laurent-indermuehle:
I was wrong. MySQL and MariaDB do differ:

```sql
-- MYSQL
+------------------+-----------+
| user             | host      |
+------------------+-----------+
| role_foo         | %         |
+------------------+-----------+

-- MariaDB
+-------------+-----------+
| User        | Host      |
+-------------+-----------+
| role_foo    |           |
+-------------+-----------+

So I close this PR. Sorry for the noise.

```

Andersson007:
@laurent-indermuehle no worries:)

Andersson007:
BTW maybe we could add this table above to the code as a comment, so that it's seen why it works that way?

PR #580, opened 2023-10-12
feat[mysql_info]: add 'users_info' filter

laurent-indermuehle:
##### SUMMARY

Add the filter `users_info` to the `mysql_info` module. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

`mysql_info`

##### ADDITIONAL INFORMATION

This filter returns information about users accounts. Useful when migrating accounts to a new server or to help exporting accounts in YAML that can be placed in an Ansible inventory.
To migrate accounts, feed the output of `mysql_info` directly to `mysql_user`.
To export the accounts definition, use a playbook that writes the output of `mysql_info` into a file or just ansible.builtin.debug if you have very few accounts.

It doesn't return information about `password_option` and `lock_option` but since the `mysql_user` module doesn't support them, I didn't implemented them.

Doesn't support `sha256_password` and `caching_sha2_password` authentication plugins. Even if you use `print_identified_with_as_hex` to convert the password into a hash. I think this is because PyMySQL and
mysqlclient wrap the password in quote: `CREATE USER... IDENTIFIED AS '0x1234'` instead of `CREATE USER... IDENTIFIED AS 0x1234`. Sorry to leave that unfinished but this issue is too much work for one who only works on MariaDB with `mysql_native_password`.


plugins/modules/mysql_info.py
```diff
@@ -122,6 +125,35 @@
...
+    column_case_sensitive: true
+    state: present
+  loop: "{{ result.users_privs }}"
+  loop_control:
+    label: "{{ item.name }}@{{ item.host }}"
+  when:
+    - item.name != 'root'  # In case you don't want to import admin accounts
+    - item.name != 'mariadb.sys'
+    - item.name != 'mysql'
+
```
```suggestion
- name: Clone users from one server to another
  block:
  # Step 1
  - name: Fetch information from a source server
    delegate_to: server_source
    community.mysql.mysql_info:
      filter:
        - users_privs
    register: result

  # Step 2
  # Don't work with sha256_password and cache_sha2_password
  - name: Clone users fetched in a previous task to a target server
    community.mysql.mysql_user:
      name: "{{ item.name }}"
      host: "{{ item.host }}"
      plugin: "{{ item.plugin | default(omit) }}"
      plugin_auth_string: "{{ item.plugin_auth_string | default(omit) }}"
      plugin_hash_string: "{{ item.plugin_hash_string | default(omit) }}"
      tls_require: "{{ item.tls_require | default(omit) }}"
      priv: "{{ item.priv | default(omit) }}"
      resource_limits: "{{ item.resource_limits | default(omit) }}"
      column_case_sensitive: true
      state: present
    loop: "{{ result.users_privs }}"
    loop_control:
      label: "{{ item.name }}@{{ item.host }}"
    when:
      - item.name != 'root'  # In case you don't want to import admin accounts
      - item.name != 'mariadb.sys'
      - item.name != 'mysql'
```
How about using a block to separate this tasks from the rest?

plugins/modules/mysql_info.py
```diff
@@ -186,6 +218,21 @@
   type: dict
   sample:
   - { "localhost": { "root": { "Alter_priv": "Y", "Alter_routine_priv": "Y" } } }
+users_privs:
+  description:
+    Information about users accounts. The output can be used as an input of the
+    mysql_user plugin. Useful when migrating accounts to a new server or to
+    create an inventory. Does not support proxy privileges. Cause issue with
+    authentications plugins C(sha256_password) and C(caching_sha2_password).
```
```suggestion
    - Information about users accounts.
    - The output can be used as an input of the M(community.mysql.mysql_user) plugin.
    - Useful when migrating accounts to another server or to create an inventory.
    - Does not support proxy privileges. Causes issues when authentications plugins C(sha256_password) and C(caching_sha2_password).
```

Andersson007 (commented):
@laurent-indermuehle thanks a lot for the PR!

>Causes issues when authentications plugins C(sha256_password) and C(caching_sha2_password).

What kind of issues? Will the module not fail? Or you mean issues for migration?
Anyway, if i have such questions, I think it's worth elaborating a bit more on the issues in the doc;)

UPDATE: Also how about renaming the key to `user_accounts`?

laurent-indermuehle:
> @laurent-indermuehle thanks a lot for the PR!
> 
> > Causes issues when authentications plugins C(sha256_password) and C(caching_sha2_password).
> 
> What kind of issues? Will the module not fail? Or you mean issues for migration? Anyway, if i have such questions, I think it's worth elaborating a bit more on the issues in the doc;)
> 
> UPDATE: Also how about renaming the key to `user_accounts`?

You're right, I was vague. I've pushed an explanation along with your suggestion of using a list in the description instead of a long sentence.

user_accounts is a far better name than users_privs (user_ or users_ thought?) I'll need to rename everything. I'll do this soon.

Andersson007:
>user_accounts is a far better name than users_privs (user_ or users_ thought?) I'll need to rename everything. I'll do this soon.

@laurent-indermuehle how about `users` if it's not occupied already? If it's occupied `users_accounts` i think
UPDATE: or maybe `users_info` if `users` is busy?

laurent-indermuehle:
> @laurent-indermuehle how about `users` if it's not occupied already? If it's occupied `users_accounts` i think UPDATE: or > maybe `users_info` if `users` is busy?

Users is already used.
I see you prefer the plural. I do too now I see it. Ok then, I re-re-name to `users_info`?
`users_information` would be too long. And `users_accounts` is kind of synonymous.

Andersson007:
`users_info` SGTM

laurent-indermuehle:
I added a description for the old `users` filter. Otherwise it was weird:

```yaml
users:
  description: Users informations
users_info:
  description: Information about users accounts...
```

If someone knows a use case for this filter, we could add it to the description.

Andersson007 (approved):
SGTM

Andersson007:
@laurent-indermuehle :tada: thanks!

PR #604, opened 2024-01-10
feat[mysql_user]: add support for mysql user attributes

n-cc:
##### SUMMARY
This PR adds support for [MySQL user attributes](https://dev.mysql.com/doc/refman/8.0/en/information-schema-user-attributes-table.html) (essentially arbitrary comments associated with a user in a key: value format) to the mysql_user module. It adds the ability to add, update, and delete user attributes, and get existing attributes by checking the `attributes` attribute of the `mysql_user` task.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`community.mysql.mysql_user`

##### ADDITIONAL INFORMATION

- Previously, `user_mod` would prematurely return at various points in check mode when changes were detected. IMO, this is not ideal, as the entirety of the module that will execute in non-check mode is not ran in check mode, and it means we have to support returning a specifically-constructed dict in multiple locations. As part of this PR, since I am modifying the return dict of `user_mod`, I removed the premature returns when in check mode in favor of the final return at the end of the function. I can revert this if desired, but I think it makes the code cleaner and safer. It is worth noting that messages from the module are still masked if a later message is registered.
- Return values are not documented yet, so I didn't document the new `attributes` return value, but I could add return value documentation if desired.

###### SERVER SUPPORTS ATTRIBUTES

Creating a new user without attributes:

```
$ ansible -m mysql_user -a '{ "user": "newuser1", "host": "localhost" }' remote
remote | CHANGED => {
    "attributes": {},
    "changed": true,
    "msg": "User added",
    "password_changed": true,
    "user": "newuser1"
}
$ 
```

Creating a new user with attributes:

```
$ ansible -m mysql_user -a '{ "user": "newuser2", "host": "localhost", "attributes": { "key": "value" } }' remote
remote | CHANGED => {
    "attributes": {
        "key": "value"
    },
    "changed": true,
    "msg": "User added",
    "password_changed": true,
    "user": "newuser2"
}
$ 
```

Updating an existing user's attributes:

```
$ ansible -m mysql_user -a '{ "user": "newuser2", "host": "localhost", "attributes": { "key": "value_changed", "new_key": "new_value" } }' remote
remote | CHANGED => {
    "attributes": {
        "key": "value_changed",
        "new_key": "new_value"
    },
    "changed": true,
    "msg": "Attributes updated: key: value_changed, new_key: new_value",
    "password_changed": false,
    "user": "newuser2"
}
$ 
```

No change when existing attributes are already in line with the attributes specified in the `attributes` parameter:

```
$ ansible -m mysql_user -a '{ "user": "newuser2", "host": "localhost", "attributes": { "key": "value_changed", "new_key": "new_value" } }' remote
remote | SUCCESS => {
    "attributes": {
        "key": "value_changed",
        "new_key": "new_value"
    },
    "changed": false,
    "msg": "User unchanged",
    "password_changed": false,
    "user": "newuser2"
}
$ 
```

No change when the `attributes` parameter is not specified:

```
$ ansible -m mysql_user -a '{ "user": "newuser2", "host": "localhost" }' remote
remote | SUCCESS => {
    "attributes": {
        "key": "value_changed",
        "new_key": "new_value"
    },
    "changed": false,
    "msg": "User unchanged",
    "password_changed": false,
    "user": "newuser2"
}
$ 
```

###### SERVER DOES NOT SUPPORT ATTRIBUTES

Creating a new user without attributes:

```
$ ansible -m mysql_user -a '{ "user": "newuser1", "host": "localhost" }' remote
remote | CHANGED => {
    "attributes": {},
    "changed": true,
    "msg": "User added",
    "password_changed": true,
    "user": "newuser1"
}
$ 
```

Reading an existing user:

```
$ ansible -m mysql_user -a '{ "user": "newuser1", "host": "localhost" }' remote
remote | SUCCESS => {
    "attributes": {},
    "changed": false,
    "msg": "User unchanged",
    "password_changed": false,
    "user": "newuser1"
}
$ 
```

Creating a new user with attributes:

```
$ ansible -m mysql_user -a '{ "user": "newuser2", "host": "localhost", "attributes": { "key": "value" } }' remote
remote | FAILED! => {
    "changed": false,
    "msg": "user attributes were specified but the mysql server does not support user attributes"
}
$ 
```

Updating an existing user's attributes:

```
$ ansible -m mysql_user -a '{ "user": "newuser1", "host": "localhost", "attributes": { "key": "value" } }' remote
remote | FAILED! => {
    "changed": false,
    "msg": "user attributes were specified but the mysql server does not support user attributes"
}
$ 
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,18 +6,6 @@ Community MySQL Collection Release Notes
 
 This changelog describes changes after version 2.0.0.
 
-v3.9.0
-======
-
-This is the minor release of the ``community.mysql`` collection.
-This changelog contains all changes to the modules and plugins in this
-collection that have been made after the previous release.
-
-Minor Changes
--------------
-
-- mysql_user - add user attribute support via the ``attributes`` parameter and return value (https://github.com/ansible-collections/community.mysql/pull/604).
-
 v3.8.0
 ======
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,18 @@ Community MySQL Collection Release Notes
 
 This changelog describes changes after version 2.0.0.
 
+v3.9.0
+======
+
+This is the minor release of the ``community.mysql`` collection.
+This changelog contains all changes to the modules and plugins in this
+collection that have been made after the previous release.
+
+Minor Changes
+-------------
+
+- mysql_user - add user attribute support via the ``attributes`` parameter and return value (https://github.com/ansible-collections/community.mysql/pull/604).
+
 v3.8.0
 ======
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -211,7 +211,7 @@ def user_add(cursor, user, host, host_all, password, encrypted,
     if tls_requires is not None:
         privileges_grant(cursor, user, host, "*.*", get_grants(cursor, user, host), tls_requires)
 
-    final_attributes = {}
+    final_attributes = None
 
     if attributes:
         cursor.execute("ALTER USER %s@%s ATTRIBUTE %s", (user, host, json.dumps(attributes)))
@@ -434,6 +434,10 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
                 module.fail_json(msg="user attributes were specified but the server does not support user attributes")
             else:
                 current_attributes = attributes_get(cursor, user, host)
+
+                if current_attributes is None:
+                    current_attributes = {}
+
                 attributes_to_change = {}
 
                 for key, value in attributes.items():
@@ -451,6 +455,9 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
                         # Final if statements excludes items whose values are None in attributes_to_change, i.e. attributes that will be deleted
                         final_attributes = {k: v for d in (current_attributes, attributes_to_change) for k, v in d.items() if k not in attributes_to_change or
                                             attributes_to_change[k] is not None}
+
+                        # Convert empty dict to None per return value requirements
+                        final_attributes = final_attributes if final_attributes else None
                     changed = True
                 else:
                     final_attributes = current_attributes
@@ -974,7 +981,6 @@ def get_attribute_support(cursor):
     Returns:
         True if attributes are supported, False if they are not.
     """
-
     try:
         # information_schema.tables does not hold the tables within information_schema itself
         cursor.execute("SELECT attribute FROM INFORMATION_SCHEMA.USER_ATTRIBUTES LIMIT 0")
@@ -994,21 +1000,16 @@ def attributes_get(cursor, user, host):
         host (str): User host name.
 
     Returns:
-        None if the user does not exist, otherwise a dict of attributes set on the user
+        None if the user does not exist or the user has no attributes set, otherwise a dict of attributes set on the user
     """
     cursor.execute("SELECT attribute FROM INFORMATION_SCHEMA.USER_ATTRIBUTES WHERE user = %s AND host = %s", (user, host))
 
     r = cursor.fetchone()
+    # convert JSON string stored in row into a dict - mysql enforces that user_attributes entires are in JSON format
+    j = json.loads(r[0]) if r and r[0] else None
 
-    if r:
-        attributes = r[0]
-        # convert JSON string stored in row into a dict - mysql enforces that user_attributes entires are in JSON format
-        if attributes:
-            return json.loads(attributes)
-        else:
-            return {}
-
-    return None
+    # if the attributes dict is empty, return None instead
+    return j if j else None
 
 
 def get_impl(cursor):
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -10,6 +10,7 @@ __metaclass__ = type
 # Simplified BSD License (see simplified_bsd.txt or https://opensource.org/licenses/BSD-2-Clause)
 
 import string
+import json
 import re
 
 from ansible.module_utils.six import iteritems
@@ -151,13 +152,19 @@ def get_existing_authentication(cursor, user, host):
 
 def user_add(cursor, user, host, host_all, password, encrypted,
              plugin, plugin_hash_string, plugin_auth_string, new_priv,
-             tls_requires, check_mode, reuse_existing_password):
+             attributes, tls_requires, reuse_existing_password, module):
     # we cannot create users without a proper hostname
     if host_all:
         return {'changed': False, 'password_changed': False}
 
-    if check_mode:
+    if module.check_mode:
         return {'changed': True, 'password_changed': None}
+
+    # If attributes are set, perform a sanity check to ensure server supports user attributes before creating user
+    if attributes and not get_attribute_support(cursor):
+        module.fail_json(msg="user attributes were specified but the mysql server does not support user attributes")
+
+    final_attributes = {}
 
     # Determine what user management method server uses
     old_user_mgmt = impl.use_old_user_mgmt(cursor)
@@ -203,9 +210,13 @@ def user_add(cursor, user, host, host_all, password, encrypted,
     if new_priv is not None:
         for db_table, priv in iteritems(new_priv):
             privileges_grant(cursor, user, host, db_table, priv, tls_requires)
+    if attributes is not None:
+        cursor.execute("ALTER USER %s@%s ATTRIBUTE %s", (user, host, json.dumps(attributes)))
+        final_attributes = attributes_get(cursor, user, host)
     if tls_requires is not None:
         privileges_grant(cursor, user, host, "*.*", get_grants(cursor, user, host), tls_requires)
-    return {'changed': True, 'password_changed': not used_existing_password}
+
+    return {'changed': True, 'password_changed': not used_existing_password, 'attributes': final_attributes}
 
 
 def is_hash(password):
@@ -218,7 +229,7 @@ def is_hash(password):
 
 def user_mod(cursor, user, host, host_all, password, encrypted,
              plugin, plugin_hash_string, plugin_auth_string, new_priv,
-             append_privs, subtract_privs, tls_requires, module, role=False, maria_role=False):
+             append_privs, subtract_privs, attributes, tls_requires, module, role=False, maria_role=False):
     changed = False
     msg = "User unchanged"
     grant_option = False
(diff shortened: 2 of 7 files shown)


plugins/modules/mysql_user.py
```diff
@@ -165,6 +164,13 @@
         fields names in privileges.
     type: bool
     version_added: '3.8.0'
+  attributes:
+    description:
+      - Create, update, or delete user attributes (arbitrary "key: value" comments) for the user.
+      - MySQL server must support the INFORMATION_SCHEMA.USER_ATTRIBUTES table. Provided since MySQL 8.0.
+      - To delete an existing attribute, set its value to False.
+    type: dict
+    version_added: '3.9.0'
```
n-cc, line 173: I did not bump the version in `galaxy.yml` yet; unsure what the process is for shipping new versions.

plugins/modules/mysql_user.py
```diff
@@ -405,6 +418,7 @@ def main():
         tls_requires=dict(type='dict'),
         append_privs=dict(type='bool', default=False),
         subtract_privs=dict(type='bool', default=False),
+        attributes=dict(type='dict'),
```
n-cc, line 421: I can change the parameter ordering here if desired.

n-cc:
I'll work on resolving the lint and integration tests over the next day or so.

n-cc:
Existing integration tests are passing, going to work on writing new ones for this feature.

n-cc:
Integration tests written, everything should be good for a review at this point.

tests/integration/targets/test_mysql_user/tasks/test_user_attributes.yml
```diff
@@ -0,0 +1,244 @@
...
+          attributes:
+            key2: "new_value2"
+        check_mode: yes
+        register: result
+
+      - name: Attributes | Assert that attribute would have been updated
+        assert:
+          that:
+            - result is changed
+            - result.attributes.key2 == "new_value2"
```
Andersson007:
```suggestion
      - name: Attributes | Assert that attribute would have been updated
        assert:
          that:
            - result is changed
            - result.attributes.key2 == "new_value2"
```
please use `mysql_query` module to get and check the actual result after each task. I wouldn't rely solely on what `mysql_user` returns. Check after each task including `check mode: true` to check that changes haven't been made, etc.

plugins/modules/mysql_user.py
```diff
@@ -257,6 +263,13 @@
       FUNCTION my_db.my_function: EXECUTE
     state: present
 
+- name: Modify user attributes, creating the attribute 'foo' and removing the attribute 'bar'
+  community.mysql.mysql_user:
+    name: bob
+    attributes:
+      foo: "foo"
+      bar: False
```
```suggestion
      bar: False
```
Could a boolean be a valid value of an attribute? (I don't really know) Could we use `None` instead?

plugins/modules/mysql_user.py
```diff
@@ -165,6 +164,13 @@
         fields names in privileges.
     type: bool
     version_added: '3.8.0'
+  attributes:
+    description:
+      - "Create, update, or delete user attributes (arbitrary 'key: value' comments) for the user."
+      - MySQL server must support the INFORMATION_SCHEMA.USER_ATTRIBUTES table. Provided since MySQL 8.0.
+      - To delete an existing attribute, set its value to False.
```
```suggestion
      - To delete an existing attribute, set its value to False.
```
In another comment, i suggested using `None`, not sure if possible though. See the comment below.

plugins/module_utils/user.py
```diff
@@ -924,6 +967,51 @@ def limit_resources(module, cursor, user, host, resource_limits, check_mode):
...
+
+    if r:
+        attributes = r[0]
+        # convert JSON string stored in row into a dict - mysql enforces that user_attributes entires are in JSON format
+        if attributes:
+            return json.loads(attributes)
+        else:
+            return {}
+
+    return None
```
```suggestion
    # convert JSON string stored in row into a dict - mysql enforces that user_attributes entires are in JSON format
    return json.loads(r[0]) if r and r[0] else None
```

plugins/modules/mysql_user.py
Andersson007: `3.9.0` is OK here

plugins/modules/mysql_user.py
Andersson007: that's fine now

plugins/module_utils/user.py
```diff
@@ -404,42 +413,76 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
...
             after_priv = privileges_get(cursor, user, host, maria_role)
             changed = changed or (curr_priv != after_priv)
 
+        # Handle attributes
+        attribute_support = get_attribute_support(cursor)
+        final_attributes = {}
+
+        if attributes:
+            if not attribute_support:
+                module.fail_json(msg="user attributes were specified but the mysql server does not support user attributes")
```
```suggestion
                module.fail_json(msg="user attributes were specified but the server does not support user attributes")
```
or if it's applicable only to `MySQL` please change it to `MySQL version`

Andersson007 (commented):
@n-cc hello, thanks for the PR!
could you please add a [changelog](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) fragment?
Also will you be responsive in case of related bug reports? If not really, should we implement it somehow separately from all those functions you're changing here? But i hope you're going to be around and even join our Matrix channel and Forum group:) See the [communication guide](https://docs.ansible.com/ansible/devel/community/communication.html).

plugins/modules/mysql_user.py
n-cc: Yeah, booleans are valid values, so using `False` to delete keys probably isn't ideal... what would be the proper way to represent `None` in a task? IIRC there were some caveats with using `null` or `None` (I thought ansible interpreted them both as strings?)

n-cc:
> Also will you be responsive in case of related bug reports? 

Certainly if there are any immediate problems, but you could also consider me a maintainer of this feature if future problems arise. I don't actively use irc/matrix, but I should be reachable via Github pings. 

That being said, if implementing attribute support separately from `user_mod` and `user_add` is preferable, I can do that, I just opted to include everything in `user_mod` as it seems like that's where all user modification is handled currently.

plugins/module_utils/user.py
n-cc: Looks like this will return `None` rather than an empty dict if the user exists but does not have any attributes set - are you recommending returning `None` over `{}` in that scenario?

plugins/modules/mysql_user.py
n-cc: Looks like `null` should work - I'll recommend the use of `null` in the documentation and update the code/tests accordingly.

n-cc:
Requested changes have been implemented, aside from the two I had questions on.

plugins/module_utils/user.py
Andersson007: yep

Andersson007:
> > Also will you be responsive in case of related bug reports?
> 
> Certainly if there are any immediate problems, but you could also consider me a maintainer of this feature if future problems arise. I don't actively use irc/matrix, but I should be reachable via Github pings.
> 
> That being said, if implementing attribute support separately from `user_mod` and `user_add` is preferable, I can do that, I just opted to include everything in `user_mod` as it seems like that's where all user modification is handled currently.

that's nice to hear! Could you please take a look at the comment about using `None`. thanks

plugins/module_utils/user.py
n-cc: That should be fine, but there are a few things to keep in mind about `None` vs `{}` when it comes to attributes... if a `CREATE USER` is ran without the `ATTRIBUTES '{}'` option, the user's attributes are stored as `NULL` until attributes are set, and if all attributes are deleted, attributes will then be stored as `{}` and cannot be changed back to `NULL`  (as far as I can tell):

```
mysql> CREATE USER 'testuser'@'localhost';
Query OK, 0 rows affected (0.00 sec)

mysql> SELECT ATTRIBUTE from INFORMATION_SCHEMA.USER_ATTRIBUTES WHERE user = 'testuser' and host = 'localhost';
+-----------+
| ATTRIBUTE |
+-----------+
| NULL      |
+-----------+
1 row in set (0.00 sec)

mysql> ALTER USER 'testuser'@'localhost' ATTRIBUTE '{ "foo": "bar" }';
Query OK, 0 rows affected (0.00 sec)

mysql> SELECT ATTRIBUTE from INFORMATION_SCHEMA.USER_ATTRIBUTES WHERE user = 'testuser' and host = 'localhost';
+----------------+
| ATTRIBUTE      |
+----------------+
| {"foo": "bar"} |
+----------------+
1 row in set (0.00 sec)

mysql> ALTER USER 'testuser'@'localhost' ATTRIBUTE '{ "foo": null }';
Query OK, 0 rows affected (0.00 sec)

mysql> SELECT ATTRIBUTE from INFORMATION_SCHEMA.USER_ATTRIBUTES WHERE user = 'testuser' and host = 'localhost';
+-----------+
| ATTRIBUTE |
+-----------+
| {}        |
+-----------+
1 row in set (0.00 sec)

mysql>
```

I think there are a couple different options:

1. Represent no attributes as `{}` even if the `ATTRIBUTE` database value is `NULL`
    1. This is what I was doing originally; I chose this method so that someone using `register: foo` with `mysql_user` could still treat `foo.attributes` as a dictionary even if the `ATTRIBUTE` value is `NULL`. It also makes the code a bit cleaner if `attributes_get` always returns a dict for users that exist.
    1. With this method, we could also explicitly run the `CREATE USER` query in `user_add` with `ATTRIBUTES {}` even if no attributes are specified, so that the `ATTRIBUTE` database value will always be `{}` for ansible-created users on servers with attribute support.
1. Represent no attributes as `None` even if the `ATTRIBUTE` database value is `{}` (I think this is what you were proposing)
1. Represent no attributes as either `None` or `{}` depending on the database value

Option 2 is how this feature behaves as of my latest commit (I also wrote tests to assert the attributes return value for users without attributes). I think any of these options are fine depending on ansible best practices. I can stick with option 2 if you think it's preferred, I just wanted to make sure this caveat was documented in this PR.

n-cc:
> > > Also will you be responsive in case of related bug reports?
> > 
> > 
> > Certainly if there are any immediate problems, but you could also consider me a maintainer of this feature if future problems arise. I don't actively use irc/matrix, but I should be reachable via Github pings.
> > That being said, if implementing attribute support separately from `user_mod` and `user_add` is preferable, I can do that, I just opted to include everything in `user_mod` as it seems like that's where all user modification is handled currently.
> 
> that's nice to hear! Could you please take a look at the comment about using `None`. thanks

Responded and updated.

CHANGELOG.rst
```diff
@@ -6,6 +6,18 @@ Community MySQL Collection Release Notes
...
+
+This is the minor release of the ``community.mysql`` collection.
+This changelog contains all changes to the modules and plugins in this
+collection that have been made after the previous release.
+
+Minor Changes
+-------------
+
+- mysql_user - add user attribute support via the ``attributes`` parameter and return value (https://github.com/ansible-collections/community.mysql/pull/604).
+
```
Andersson007:
```suggestion
```
this will be generated automatically during the next release based on a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment) you need to add. Please remove this and add the fragment.

Andersson007 (changes requested):
@n-cc one small thing to change, the rest LGTM

CHANGELOG.rst
n-cc: My bad, should've read the documentation closer

n-cc:
> @n-cc one small thing to change, the rest LGTM

Thanks much, updated

Andersson007 (approved):

Andersson007:
@laurent-indermuehle is it OK for you? if yes, please merge

Andersson007:
as @laurent-indermuehle is busy, I'm merging this PR

@n-cc thanks for the contribution!

PR #623, opened 2024-04-08
Add tests for Ansible core 2.17 (devel is 2.18 today) and bump tests dependencies

### LINKED ISSUE #622: GHA: add testing against stable-2.17
##### SUMMARY
As stable-2.17 of ansible-core has been recently [branched](https://forum.ansible.com/t/ansible-core-devel-bumped-to-2-18-0-dev0-stable-2-17-branch-created/4695/1) (so now devel is 2.18), we need to explicitly add testing against `stable-2.17` to our GHA CI matrix

laurent-indermuehle:
##### SUMMARY
fix #622

Drop tests for Ansible-core 2.14 and add 2.17.

I had to upgrade from Ubuntu 20.04 to 22.04 to overcome the error:
```
ERROR: Command "docker run --tmpfs /tmp:exec --tmpfs /run:exec --tmpfs /run/lock --volume /var/run/docker.sock:/var/run/docker.sock --cgroupns host --tmpfs /sys/fs/cgroup --volume /sys/fs/cgroup/systemd:/sys/fs/cgroup/systemd:rw -dt --ulimit nofile=10240 --name ansible-test-controller-jP3nWvru --network github_network_aaa99831959740bca775ef73f797b985 '  '" returned exit status 125.
``` 
As a reminder, ubuntu 20.04 was forced because cgroup v2 wasn't supported by ansible 2.10, 2.11 and 2.12 (more info [here](https://github.com/ansible-collections/collection_template/blob/main/.github/workflows/ansible-test.yml#L83-L91). Since we dropped those versions we can now move on to the latest ubuntu.
Also latest ansible-test-gh-action doesn't work with ubuntu 20.04, so the upgrade is necessary.

I also updated some GitHub Action plugins to avoid warnings. But some warnings comes from ansible-community/ansible-test-gh-action@release/v1 which as this PR that will solve them when merged: https://github.com/ansible-community/ansible-test-gh-action/pull/80

Finally, I disabled the role tests because they were out of date and we don't use them anyway.

##### ISSUE TYPE
- Github action - Integration tests

##### COMPONENT NAME

- Github action - Integration tests

### DIFF
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -1,6 +1,6 @@
 ---
 name: Plugins CI
-on:
+on:  # yamllint disable-line rule:truthy
   push:
     paths:
       - 'plugins/**'
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -275,8 +275,8 @@ jobs:
           fi
 
       - name: Set docker_image
-        run: |
-          docker_image_multiline=(" \
+        run: |-
+          docker_image_multiline=("\
           ghcr.io/ansible-collections/community.mysql\
           /test-container-${{ env.db_client }}\
           -py${{ env.python_version_flat }}\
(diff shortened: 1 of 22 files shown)


Andersson007 (approved):

Andersson007:
@laurent-indermuehle thanks for the great job you do and for contributing long-term!

PR #365, opened 2022-05-19
mysql_user: add "update_password: on_new_username" argument, "password_changed" result field

betanummeric:
##### SUMMARY
fixes #344

- Add value `on_new_username` to `update_password` argument of module mysql_user.
- Add result field `password_changed` boolean to module mysql_user.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module mysql_user

##### ADDITIONAL INFORMATION

When `update_password: on_new_username` is set, reuse the password (plugin and authentication_string) when creating a new user if some user with the same name already exist. If the existing users with the same name have varying passwords, the password from the arguments is used like with `update_password: always`.

The result field `password_changed` is true, when the user got a new password. When the user was created with `update_password: on_new_username` and an existing password was reused, `password_changed` is false.

Issue #344, opened 2022-05-10
mysql_user: add option to reuse password of existing user when creating a new one


betanummeric:
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
mysql_user

##### SUMMARY
In MySQL/MariaDB the password is a property of a username-source-address-combination. However, I want to use the same password per username, for all source addresses of that username. (I only need password authentication, I'm not sure how this generalizes to other authentication methods).
For that, I would need to reuse the existing password when creating a new user with the same name. This new behavior could be enabled with a new value for the existing module argument: `update_password: on_new_username`.

For further processing, I need a `password_changed` boolean in the module return object. This could/should also be used to indicate whether the password of an existing user was changed.

I can submit a PR for this.

##### EXAMPLE

When I request a new user with
```yaml
- name: add user with other IP
  community.mysql.mysql_user:
    state: present
    name: "test"
    host: "2001:db8::1"
    password: "newpassword"
    update_password: on_new_username
```

- and the user `"test"@"localhost"` already exists with password `existingpassword`, I would like to get the user `"test"@"2001:db8::1"` with password `existingpassword` (and ignore the `newpassword`). `password_changed` should be `false`.
- and there are multiple preexisting users with the same name, but varying passwords, I suggest to fall back to the supplied `newpassword`. `password_changed` should be `true`.
- and there is no user called `test`, the `"test"@"2001:db8::1"` user is created normally with the password `newpassword`. `password_changed` should be `true`.

hubiongithub:
Hello @betanummeric 

Can't this be solved outside of mysql.user by using items and adding hosts to the list as required:

```
- name: add user with other IP
  community.mysql.mysql_user:
    state: present
    name: "test"
    host: {{ item }}
    password: "password you want for these users"
    update_password: on_create/always       # depending on wether ansible is the truth or the database over time
  with_items:
    - host1
    - host2
    - ...
```

This works for all kind of auth methods and no need to read the hashes ...

betanummeric:
Hi @hubiongithub, my point is that I want to reuse a password/hash which only exists on the mysql/mariadb instance - I don't know it in the inventory or playbook. (The truth lies in the database, not in ansible.) Of course I can query the existing hash outside of the mysql_user module and pass it there with `update_password: on_create`, but I thought it would be better placed in the module.

My PR just looks for an existing `plugin`-`authentication_string`-combination, so it should work with all kinds of authentication methods.

[...]

PR #629, opened 2024-04-12
Fix hashed passwords being returned by get_existing_authentication() via the plugin_auth_string variable instead of plugin_hash_string

### LINKED ISSUE #621: `mysql_user`: can't get `plugin_hash_string` working with `caching_sha2_password`
##### SUMMARY
Hello,
I use `caching_sha2_password` plugin for my users on MySQL 8 servers.
At beginning, I just use 
```yaml
- name: myyser
  community.mysql.mysql_user:
    name: myuser
    plugin_auth_string: "xxxxx"
    plugin: "caching_sha2_password"
```
but I had idempotence issue because, each time the password was marked as changed because the hash generated by MySQL was different.
So I tried to use `plugin_hash_string` instead of `plugin_auth_string` to provide directly the hash generated from ansible to avoid idempotence issue. Like
```yaml
- name: myyser
  community.mysql.mysql_user:
    name: myuser
    plugin_hash_string:  "$A$005$9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0fd02fe"
    plugin: "caching_sha2_password"
```
but I have many problems:
1. [the test done on this ansible module is wrong](https://github.com/ansible-collections/community.mysql/blob/bfe2fdc3ff8b94b14574cdade1639ce11877215c/tests/integration/targets/test_mysql_info/tasks/filter_users_info.yml#L210) it should use `plugin_hash_string` instead of `plugin_auth_string`
2. As hash generated by MySQL are not string and contains special character, I can't provide a correct hash from ansible, maybe I can initiate a PR to provide a base64 encoded hash that will be decoded before send to MySQL ? Like `plugin_hash_string = base64.b64decode(plugin_hash_base64).decode("utf-8")`
3. The lack of documentation of how generate a correct hash is a problem
I found [this](https://github.com/mysql/mysql-server/blob/ea7d2e2d16ac03afdd9cb72a972a95981107bf51/sql/auth/sha2_password.cc#L397) on MySQL source code, with the hash like `$A$005$salthash` (salf length 20, hash length 43) but cannot log with it. Maybe a correct example can be added to the ansible documentation ?

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
mysql_user

##### ANSIBLE VERSION
```ansible [core 2.16.2]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.11/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION
```Collection      Version
--------------- -------
community.mysql 3.9.0  
----------------- -------
community.general 8.5.0  
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT
Debian 12

##### STEPS TO REPRODUCE
example of hash
```yaml
- name: myyser
  community.mysql.mysql_user:
    name: myuser
    plugin_hash_string: "{{ '$A$005$' ~ 'abcdefghtrdlsorafgthe' ~ ('test' | hash('sha256'))[0:42] }}"
    plugin: "caching_sha2_password"
```

##### EXPECTED RESULTS
Changed hash, can connect with password before hash, and idempotence ok

##### ACTUAL RESULTS
Change hash, cannot connect with password
or
Hash not accepted by MySQL

laurent-indermuehle:
##### SUMMARY

Fix hashed passwords being returned by `get_existing_authentication()` via the `plugin_auth_string` variable instead of `plugin_hash_string`. This method is used in `user_add()`, which doesn't seem appropriate. I would have expected to find it in `user_mod()` instead. I hope it is never used and doesn't require to treat this PR as a major_change.
The only other place where `get_existing_authentication()` is used is in `mysql_info` while utilizing the `users_info` filter introduced in 3.8.0 (#580). Since this feature is relatively new, I don't expect it to have widespread usage yet.

This doesn't f-i-x #621 but the issue was reported there. It makes more sense to use the right variable anyway.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

mysql_user, mysql_info

##### ADDITIONAL INFORMATION

Using `mysql_info` with community.mysql 3.9.0:
```yaml
users_info:
    - host: localhost
      name: root
      plugin: mysql_native_password
      plugin_auth_string: '*7D2ABFF56C15D67445082FBB4ACD2DCD26C0ED57'
      priv: '*.*:ALL,GRANT'
```

Using `mysql_info` with this PR:
```yaml
users_info:
    - host: localhost
      name: root
      plugin: mysql_native_password
      plugin_hash_string: '*7D2ABFF56C15D67445082FBB4ACD2DCD26C0ED57'  # <------ Fixed
      priv: '*.*:ALL,GRANT'

```

### DIFF
--- a/changelogs/fragments/lie_fix_plugin_hash_string_return.yml
+++ b/changelogs/fragments/lie_fix_plugin_hash_string_return.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - mysql_info - Fix ``users_info`` filter output variable for hashed password to use ``plugin_hash_string`` instead of ``plugin_auth_string`` (https://github.com/ansible-collections/community.mysql/pull/629).
--- a/changelogs/fragments/lie_fix_plugin_hash_string_return.yml
+++ b/changelogs/fragments/lie_fix_plugin_hash_string_return.yml
@@ -1,3 +1,6 @@
 ---
 bugfixes:
   - mysql_info - Add ``plugin_hash_string`` to ``users_info`` filter's output. The existing ``plugin_auth_string`` contained the hashed password and thus is missleading, it will be removed from community.mysql 4.0.0. (https://github.com/ansible-collections/community.mysql/pull/629).
+
+breaking_changes:
+  - mysql_info - The ``users_info`` filter returned variable ``plugin_auth_string`` contains the hashed password and it's misleading, it will be removed from community.mysql 4.0.0. Use the `plugin_hash_string` return value instead (https://github.com/ansible-collections/community.mysql/pull/629).
--- a/changelogs/fragments/lie_fix_plugin_hash_string_return.yml
+++ b/changelogs/fragments/lie_fix_plugin_hash_string_return.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - mysql_info - Fix ``users_info`` filter output variable for hashed password to use ``plugin_hash_string`` instead of ``plugin_auth_string`` (https://github.com/ansible-collections/community.mysql/pull/629).
+  - mysql_info - Add ``plugin_hash_string`` to ``users_info`` filter's output. The existing ``plugin_auth_string`` contained the hashed password and thus is missleading, it will be removed from community.mysql 4.0.0. (https://github.com/ansible-collections/community.mysql/pull/629).
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -118,11 +118,21 @@ def get_existing_authentication(cursor, user, host):
     if isinstance(rows, dict):
         rows = list(rows.values())
 
+    # 'plugin_auth_string' contains the hash string. Must be removed in c.mysql 4.0
+    # See https://github.com/ansible-collections/community.mysql/pull/629
     if isinstance(rows[0], tuple):
-        return {'plugin': rows[0][0], 'plugin_hash_string': rows[0][1]}
+        return {'plugin': rows[0][0],
+                'plugin_auth_string': rows[0][1],
+                'plugin_hash_string': rows[0][1]
+        }
 
+    # 'plugin_auth_string' contains the hash string. Must be removed in c.mysql 4.0
+    # See https://github.com/ansible-collections/community.mysql/pull/629
     if isinstance(rows[0], dict):
-        return {'plugin': rows[0].get('plugin'), 'plugin_hash_string': rows[0].get('auth')}
+        return {'plugin': rows[0].get('plugin'),
+                'plugin_auth_string': rows[0].get('auth'),
+                'plugin_hash_string': rows[0].get('auth')
+        }
     return None
 
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -123,16 +123,14 @@ def get_existing_authentication(cursor, user, host):
     if isinstance(rows[0], tuple):
         return {'plugin': rows[0][0],
                 'plugin_auth_string': rows[0][1],
-                'plugin_hash_string': rows[0][1]
-        }
+                'plugin_hash_string': rows[0][1]}
 
     # 'plugin_auth_string' contains the hash string. Must be removed in c.mysql 4.0
     # See https://github.com/ansible-collections/community.mysql/pull/629
     if isinstance(rows[0], dict):
         return {'plugin': rows[0].get('plugin'),
                 'plugin_auth_string': rows[0].get('auth'),
-                'plugin_hash_string': rows[0].get('auth')
-        }
+                'plugin_hash_string': rows[0].get('auth')}
     return None
 
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -162,7 +162,7 @@ def user_add(cursor, user, host, host_all, password, encrypted,
         existing_auth = get_existing_authentication(cursor, user, host)
         if existing_auth:
             plugin = existing_auth['plugin']
-            plugin_hash_string = existing_auth['auth_string']
+            plugin_hash_string = existing_auth['plugin_hash_string']
             password = None
             used_existing_password = True
     if password and encrypted:
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -143,10 +143,10 @@ def get_existing_authentication(cursor, user, host):
         rows = list(rows.values())
 
     if isinstance(rows[0], tuple):
-        return {'plugin': rows[0][0], 'plugin_auth_string': rows[0][1]}
+        return {'plugin': rows[0][0], 'plugin_hash_string': rows[0][1]}
 
     if isinstance(rows[0], dict):
-        return {'plugin': rows[0].get('plugin'), 'plugin_auth_string': rows[0].get('auth')}
+        return {'plugin': rows[0].get('plugin'), 'plugin_hash_string': rows[0].get('auth')}
     return None
 
 
--- a/tests/integration/targets/test_mysql_info/tasks/filter_users_info.yml
+++ b/tests/integration/targets/test_mysql_info/tasks/filter_users_info.yml
@@ -186,7 +186,7 @@
           - users_info
       register: result
 
-    - name: Recreate users from mysql_info users_info result
+    - name: Mysql_info users_info | Recreate users from mysql_info result
       community.mysql.mysql_user:
         name: "{{ item.name }}"
         host: "{{ item.host }}"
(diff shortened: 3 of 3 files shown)


Andersson007 (approved):
@laurent-indermuehle thanks for the PR! Makes sense to change, though looks like a breaking change as we're changing the interface here.
Maybe it's time to plan the major release and incorporate it in https://github.com/ansible-collections/community.mysql/milestone/1 ?

For now, can we have both the keys having the same content (i.e. change your PR to have both)
if it's possible, we could add a deprecation message to the changelog under major_changes/breaking changes now saying that the wrong one will go away in 4.0.0.
In several month we could release 4.0.0 - it'll be at least a bit expected for users.

I'm not insisting though. Please decide how you think it'll be OK and I'll share the responsibility in case of any complaints (which might not even happen, not sure if anyone checks that key in playbook conditions).

laurent-indermuehle:
@Andersson007 I'm lost. The original method `get_existing_authentication` returns `plugin` and `plugin_auth_string`. But then... can you explain this line: https://github.com/ansible-collections/community.mysql/blob/main/plugins/module_utils/user.py#L155 ? Where `auth_string` comes from?

We should be more careful when we add a functionality that is not supported by MySQL and MariaDB. It adds complexity for no obvious reasons. For instance, I've read https://github.com/ansible-collections/community.mysql/issues/344 and feel like it could have been solved in the playbook with 1-2 tasks and loop. I thank @betanummeric for this feature that adds to the value of c.mysql. But it is now hard to maintain.

All the tests are passing before and after I change the variables. I've lost my confidence about this change.
I think it's because in user_add() and user_mod() `plugin_hash_string` has a higher precedence over `plugin_auth_string`.

But yes, you're right. I can add `plugin_hash_string` and keep `plugin_auth_string` as before. I'm changing my PR accordingly. I push to see if CI is all green, but I'm unhappy with the complexity of the password handling.

changelogs/fragments/lie_fix_plugin_hash_string_return.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - mysql_info - Add ``plugin_hash_string`` to ``users_info`` filter's output. The existing ``plugin_auth_string`` contained the hashed password and thus is missleading, it will be removed from community.mysql 4.0.0. (https://github.com/ansible-collections/community.mysql/pull/629).
```
Andersson007, line 3:
```suggestion
  - mysql_info - Add ``plugin_hash_string`` to ``users_info`` filter's output. The existing ``plugin_auth_string`` contained the hashed password and thus is missleading, it will be removed from community.mysql 4.0.0. (https://github.com/ansible-collections/community.mysql/pull/629).
  
 breaking_changes:
 - mysql_info - The existing ``plugin_auth_string`` return value contains the hashed password and it's missleading, it will be removed from community.mysql 4.0.0. Use the `plugin_hash_string` return value instead (https://github.com/ansible-collections/community.mysql/pull/629).
```

Andersson007 (commented):
@laurent-indermuehle let's emphasize the (potentially) breaking change in the changelog.

Andersson007:
@betanummeric any concerns on this PR?

betanummeric:
Oh, it looks like the "update_password: on_new_username" feature of `mysql_user` which was introduced by https://github.com/ansible-collections/community.mysql/pull/365 (@laurent-indermuehle `auth_string` comes from there) was broken by https://github.com/ansible-collections/community.mysql/pull/580, so since v3.8.0. :(

The `on_new_username` feature attempts to reuse an existing password when the username matches, even if the host of the user doesn't match.
For example, say the user `foo@'192.0.2.1'` already exists with password `asdf`.
When I execute `mysql_user` to create a new user `foo@'192.0.2.100'`, `mysql_user` with `update_password: on_new_username`  will set the password of the exiting `foo@'192.0.2.1'` also for `foo@'192.0.2.100'`, regardless of what new password I specified in the module arguments.
Then I manually set a different password, only for `foo@'192.0.2.100'`.
When I then create a third user `foo@'192.0.2.99'`, also with `update_password: on_new_username`, reusing the existing password will not work because there are multiple distinct options. In this case the password from the module arguments is used for the new user `foo@'192.0.2.99'`.

The feature is currently broken because
- when the user does not already exists, an `IndexError` will be raised here: https://github.com/ansible-collections/community.mysql/blob/main/plugins/module_utils/user.py#L121
- The host part must be ignored, it should check for all users with the same name: https://github.com/ansible-collections/community.mysql/blob/main/plugins/module_utils/user.py#L152
- The `get_existing_authentication` function should only return something when there is exactly one distinct password/authentication_string, to avoid ambiguity. (Therefore the `limit 2` in the SQL).

I am using this feature in production so I am happy I did not upgrade to the broken version yet and I would be glad if we can fix this.

Sure, it would be possible to work around this outside of the module, but I thought others could use this feature too. I think the logic is simple enough to maintain, but apparently the documentation could have been better.

laurent-indermuehle:
@betanummeric thanks for your reply.
I'm sorry if my PR #580 broke something. To be honest, when I wrote it I was unaware of the difference between MySQL and MariaDB for handling roles and also the `update_password` feature. 

But, if the tests still passes, it means they are incomplete! We should fix that.

laurent-indermuehle:
@betanummeric could you confirm something please?
When reading the documentation:
```yaml
- "C(on_new_username) works like C(on_create), but it tries to reuse an existing password: If one different user
        with the same username exists, or multiple different users with the same username and equal C(plugin) and
        C(authentication_string) attribute, the existing C(plugin) and C(authentication_string) are used for the
        new user instead of the I(password), I(plugin), I(plugin_hash_string) or I(plugin_auth_string) argument."
```
It's not clear to me what will happen if I have 2 accounts with same username but **different** passwords and I'm creating a 3rd account with `update_password: on_new_username`. And if I can leave the password unset.

The integration tests [here](https://github.com/ansible-collections/community.mysql/blob/main/tests/integration/targets/test_mysql_user/tasks/test_update_password.yml#L122) suggests that the plugin will use the password set in the arguments.

But I thought this feature would be used without setting any password at all. Since you explicitly want to reuse an existing one? I was expecting a failure if the account already exists with different passwords.

When I will understand this, I will be able to decide what to do with the current PR. Let me know if you feel that this PR could be merged and we fix `update_password` in another one.

betanummeric:
@laurent-indermuehle 
> It's not clear to me what will happen if I have 2 accounts with same username but different passwords and I'm creating a 3rd account with `update_password: on_new_username`.

In that case the module arguments `password`, `plugin`, `plugin_hash_string` and `plugin_auth_string` are applied to the user as if it had been `update_password: on_create`.

I intentionally built `on_new_username` as a best-effort attempt which will fallback to `on_create` if it finds multiple different passwords. Changing that to fail upon ambiguity would be a breaking change.

laurent-indermuehle:
@betanummeric thank you for confirming the first part and the explanation about `limit 2`.

I get that the root cause is that `get_existing_authentication()` only searched for usernames and in v3.8.0 I added the hostname as well. So in the tests for `update_password: on_new_username`, `user_add()` search for an existing password, doesn't found any, so it apply `test_password3`. Which, by accident, is what the test expects.

So, I think we need to:

- Add a test when no previous user exists (IndexError)
- Refactor get_existing_authentication() to support both `host` and `retrurn_all_hosts`. Or split this into 2 methods `get_user_host_existing_authentication()` and `get_user_existing_authentation()`.
- Add a warning/failure if multiple accounts with same username but different passwords exists

Will that works?

betanummeric:
- When no previous user exists, `update_password: on_new_username` should set the password from the arguments without error. Yes, there should be a test for that.
- I would make `get_existing_authentication` support both usecases, but splitting it into two functions would work as well.
- We can add a warning when multiple different passwords are found, but please no failure.

Andersson007:
any updates on this? there's another PR depending on this one

laurent-indermuehle:
I'm busy on something else. I won't have time before few weeks. @betanummeric have you some time to add the missings tests?

laurent-indermuehle:
Ok @betanummeric I know what's wrong... the tasks file `test_update_password.yml` that contains your tests is not included in main.yml... Just by adding that include, I can see the IndexError!

So, I'm going to merge this PR that fix another real issue. And open a new one that address the regression!

Andersson007:
@laurent-indermuehle +1

PR #634, opened 2024-04-29
Improve get replica/primary status

laurent-indermuehle:
##### SUMMARY

Getting the Replica/Primary status rely on testing that `cursor.fetchone()` returned a dict. To prevent case were fetchone returns a dict also in case of failure, I added tests for keys inside the dictionary.

For **getreplica**, since MariaDB and MySQL doesn't have the same variables, I searched for both Slave_IO_Running and Replica_IO_Running.

for **getprimary**, the output would be the same on a primary or a replica with binary log enabled. This is confusing with the term `is_primary` so I added "Meaning: Binary logs are disabled" to the returned message.
I searched for "file" and "Position" in the output that is luckily the same for MariaDB and MySQL:
```
mysql> show master status\G
*************************** 1. row ***************************
             File: binlog.000003
         Position: 157
     Binlog_Do_DB: 
 Binlog_Ignore_DB: 
Executed_Gtid_Set: 
1 row in set (0.00 sec)
```

This PR changes nothing and may even break in the future is MySQL or MariaDB changes the term used in the output of the command `show replica status` or `show master status`. But this will makes us aware of the change.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

mysql_replication

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/improve_get_replica_primary_status.yml
+++ b/changelogs/fragments/improve_get_replica_primary_status.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+
+  - mysql_replication - Improve detection of IsReplica and IsPrimary by inspecting the dictionary returned from the SQL query instead of relying on variable types. This ensures compatibility with changes in the connector or the output of SHOW REPLICA STATUS and SHOW MASTER STATUS, allowing for easier maintenance if these change in the future.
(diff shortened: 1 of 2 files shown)


Andersson007 (approved):

Andersson007:
@laurent-indermuehle thanks for the improvement!

PR #643, opened 2024-06-06
Embed `pymysql` connector within the collection

laurent-indermuehle:
## SUMMARY
This PR introduces significant changes to the MySQL connector integration. It's aimed at streamlining dependencies and ensuring consistent behavior by removing mysql_client support.

pymysql has a MIT license, who is compatible with the GPL-v3.

I believe this change will help us create better modules with better documented returned values and offer a simpler experience for our users. A significant portion of exceptions from our modules documentation could be removed.

## Changes
- Embed pymysql connector within the collection: This removes the need for external dependencies.
- Remove custom test containers: Leverage the default containers from ansible-test for simplicity.
- Expand test matrix: Add tests against Python 3.11 and 3.12.
- Simplify codebase: Eliminate logic related to connector presence and version checks.

## Impact on Users
If you are a user of mysql_client, please note that this change forces the usage of pymysql exclusively. Be aware that:

- Behavioral Changes: The behavior of some tasks may change, as pymysql and other connectors sometimes produce different outputs for the same actions. The most notable difference could be task results showing as green instead of yellow or vice versa.

## Pro
- Reduced Dependencies: No need to install pymysql on each controlled node (no more pip install pymysql or dnf install pymysql).
- Consistent Behavior: Ensures predictable behavior across all environments, eliminating confusion over which connector is used and for which version.
- Simplified Testing: Fewer tests are needed, reducing maintenance overhead.

## Cons
Thanks to @betanummeric, @Andersson007 and @markuman we also have a list of downside to look out for:

- We need to release the collection each time PyMySQL issue a security fix. So the CI will need new tool to help us maintain the connector (dependabot?).
- The collection may be slower because PyMySQL will need to be passed via SSH to the controlled node (up to 162Ko of Python)
- Remove the choice of the connector from our users
- Adds a breaking change

## Request for Feedback
Given the scope of these changes, your feedback is crucial. Especially if you utilize mysql_client.


laurent-indermuehle:
@betanummeric, @rsicart, @Andersson007 and any other contributor. I would love your input on the idea of embedding the pymysql connector into the collection.

betanummeric:
Please don't make pymysql part of this repository. It should be installed separately, we don't need to become pymysql maintainers.

Also including pymysql would slow down module execution a bit because pymysql needs to be copied from the ansible controller to the target node. So far, pymysql is pre-installed on the target node and doesn't need to travel through the ansible ssh connection.

What if someone wants to use mysql_client? Why is pymysql better?

"Expand test matrix: Add tests against Python 3.11 and 3.12." <- please open separate PRs for independent changes

laurent-indermuehle:
This PR as been rejected by almost all maintainers. The main reasons are security concerns  (must release c.mysql everytime PyMySQL receive a security patch) and speed (Ansible must send the whole PyMySQL code through SSH each time a module is used in a Playbook).

I extracted the part were I removed the custom test containers and open PR #650.

Also, there is a vote to deprecate mysqlclient and support only PyMySQL in #645.

For the above reasons, I close this PR. Thank you all for your feedback.

Andersson007:
It was a good and productive discussion here and on Matrix, thanks everyone!

PR #648, opened 2024-06-18
Update README.md

Andersson007:
##### SUMMARY
This badge is broken probably because we don't test roles


laurent-indermuehle (approved):
We don't have roles, this badge was misleading. Better to remove it. Thanks @Andersson007!

Andersson007:
@laurent-indermuehle agreed, thanks for reviewing!

PR #649, opened 2024-06-18
mysql_info: add server_engine return value

### LINKED ISSUE #644: Feature request: Add engine option to the mysql_info module?
Currently `community.mysql.mysql_info` doesn't have an output for the database engine that is in use, MariaDB or MySQL, there are times when this is needed, for example the [Matomo](https://matomo.org/) database schema need to be set to `Mysql` or `Mariadb`:

```bash
php console config:get --section database --key schema
"Mariadb"
```

Currently I use this to get the schema from the database:

```yaml
    - name: "Check the database version using the details from {{ matomo_db_config }}"
      community.mysql.mysql_info:
        config_file: "{{ matomo_db_config }}"
        filter:
          - version
      register: matomo_mysql_info
  
    - name: Debug matomo_mysql_info
      ansible.builtin.debug:
        var: matomo_mysql_info
        verbosity: "{% if ansible_check_mode | bool or ansible_diff_mode | bool %}1{% else %}2{% endif %}"
  
    - name: Set a fact for the database type
      ansible.builtin.set_fact:
        matomo_db_type: "{{ matomo_mysql_info.version.suffix | ansible.builtin.regex_replace('-.*$') | capitalize | trim }}"  
```

And this to set it:

```yaml
    - name: Update the Matomo database schema
      ansible.builtin.command: "{{ matomo_php }} console --no-ansi -n config:set '{{ matomo_set_database_schema }}'"
      args:
        chdir: "{{ matomo_html }}"
      vars:
        matomo_set_database_schema: 'database.schema="{{ matomo_db_type }}"'
      become: true
      become_user: "{{ matomo_user }}"
      changed_when: true
```

It would be nicer if `mysql_info` returned something like this (`engine` has been added):

```yaml
    matomo_mysql_info:
        changed: false
        connector_name: pymysql
        connector_version: 1.0.2
        failed: false
        version:
            full: 10.11.6-MariaDB-0+deb12u1
            major: 10
            minor: 11
            release: 6
            suffix: MariaDB-0+deb12u1
            engine: MariaDB
```

So I could omit the `regex_replace`.

Andersson007:
##### SUMMARY
Fixes https://github.com/ansible-collections/community.mysql/issues/644

To me it seems OK as is but I'm not sure 100% if
- it's OK now
- or we should choose another name for a key instead of `server_engine`
- or/and if we should properly capitalize the returned values, i.e. mysql -> MySQL and mariadb -> MariaDB

Thoughts? If you're OK, just approve the PR

### DIFF
--- a/changelogs/fragments/1-mysql_info.yml
+++ b/changelogs/fragments/1-mysql_info.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- mysql_info - return a database server engine used (https://github.com/ansible-collections/community.mysql/issues/644).
--- a/plugins/modules/mysql_info.py
+++ b/plugins/modules/mysql_info.py
@@ -162,6 +162,12 @@ EXAMPLES = r'''
 '''
 
 RETURN = r'''
+server_engine:
+  description: Database server engine.
+  returned: if not excluded by filter
+  type: str
+  sample: 'mariadb'
+  version_added: '3.10.0'
 version:
   description: Database server version.
   returned: if not excluded by filter
@@ -765,6 +771,7 @@ def main():
     mysql = MySQL_Info(module, cursor, server_implementation, user_implementation)
 
     module.exit_json(changed=False,
+                     server_engine=server_implementation,
                      connector_name=connector_name,
                      connector_version=connector_version,
                      **mysql.get_info(filter_, exclude_fields, return_empty_dbs))
--- a/plugins/modules/mysql_info.py
+++ b/plugins/modules/mysql_info.py
@@ -166,7 +166,7 @@ server_engine:
   description: Database server engine.
   returned: if not excluded by filter
   type: str
-  sample: 'mariadb'
+  sample: 'MariaDB'
   version_added: '3.10.0'
 version:
   description: Database server version.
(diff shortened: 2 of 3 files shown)


Andersson007:
ready for review

chriscroome:
> we should choose another name for a key instead of `server_engine`

The best I could think of was just `engine`, `server_engine` is fine, `backend` or `server_backend` might work :shrug: are there other examples in Ansible we could copy?

> or/and if we should properly capitalize the returned values, i.e. mysql -> MySQL and mariadb -> MariaDB

I'd suggest it would be best to properly capitalize them as it is easy for people to use the Jinja2 filters, to [capitalize](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.capitalize), [lower](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.lower) and [upper](https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.upper) but there is no filter that could produce `MySQL` or `MariaDB` from one of the other cases.

markuman (approved):

Andersson007:
@chriscroome SGTM, changed, PTAL

laurent-indermuehle:
I often use `db_engine`.
This name is used in the test matrix: https://github.com/ansible-collections/community.mysql/blob/main/.github/workflows/ansible-test-plugins.yml#L49-L57

`server_engine` is fine too. We also have a `server_implementation` method. So maybe it's best to stick with the same prefix?

I believe @chriscroome is right that using the correct capitalization up front makes thing easier (upper, lower, ..).

laurent-indermuehle (approved):

Andersson007:
@chriscroome thanks for requesting the feature and reviewing!
@markuman @laurent-indermuehle thanks for reviewing!
As there's no strong opinion, leaving the naming as-is, thanks everyone for your input!

PR #650, opened 2024-06-20
[CI] Remove ansible-test custom containers

laurent-indermuehle:
##### SUMMARY
This pull request aims to simplify our testing setup by removing custom test containers and using the default container provided by `ansible-test`.

##### Background
The current setup requires building custom test containers and cleaning up the registry manually, which is complex, especially for those not familiar with the system's initial implementation.

##### Changes Introduced
* Removal of Custom Test Containers
* Adoption of Default Ansible-Test Container:
  * Locally, the tests will run on an Ubuntu 20.04 container.
  * In GitHub Actions (GHA), the tests will run on an Ubuntu 22.04, utilizing `ansible-community/ansible-test-gh-action`.

##### Important Notes
* The main goal of pull request #490, which introduced custom `ansible-test` containers, was to use MySQL/MariaDB containers instead of `dbinstaller`. This PR does not alter that setup.

##### Benefits
* **Reduced Complexity**: Simplifies the process of adding/removing versions of the components.
* **Ease of Onboarding**: Makes it easier for new contributors to understand and work with the test setup.
* It's not longer to install the .deb and .py packages on the default container than using a custom container.
* The test target `controller_setup` clearly states what is installed in the test container. No need to check the Containerfile in the hidden folder `.github`. Also, this allows using Ansible instead of bash, again simplifying things.

##### Risks
I believe the difference in Ubuntu images is a non-issue. We don't test Ubuntu here, and as long as all the test components are working in both environments, the job is done.
I don't see an easy way to align versions used:
* `make` starts `ansible-test` with `--docker` from a Python venv. If the container version is forced using `--target docker:ubuntu2204`, then Python 3.10 is forced too. With the `default` container, all versions of Python are available (2.7 to 3.12 as of today).
* GHA uses Ubuntu 22.04 explicitly and then starts `ansible-test` **without** the `--docker` option.
Unless someone sees a way to make this more consistent between the two test methods, I propose to let this be as is and deal with it again when `ansible-test` changes its default container.

##### ISSUE TYPE

- simplification

##### COMPONENT NAME

- CI


Andersson007:
yeah, it'd be great if we use that default container only:)

laurent-indermuehle:
Green, finally!
All integrations tests took less than 10 minutes, so not longer than before!
Ready for review if someone could please have a look?

Andersson007 (approved):
@laurent-indermuehle looks like a great simplification! Thanks! You're a CI grand-guru:)

Andersson007:
After we drop mysqlclient it'll get significantly faster, looking forward to it

laurent-indermuehle:
@Andersson007 Yes and no, thanks to parallel execution on GHA. But we will use less resources for sure.
Also, Can I have a badge `CI grand-guru` for my resume? :P

Andersson007:
> @Andersson007 Yes and no, thanks to parallel execution on GHA. But we will use less resources for sure. Also, Can I have a badge `CI grand-guru` for my resume? :P

sure, if you create it you can:)

PR #660, opened 2024-07-17
[CI] Add 2024 versions to tests

### LINKED ISSUE #624: GHA: add testing against Python 3.11, 3.12 and PyMySQL versions present in RHEL8 and RHEL9
##### SUMMARY
Currently the tests matrix doesn't match the available packages of PyMySQL for RHEL8 and RHEL9.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

Tests

##### ADDITIONAL INFORMATION

I propose to add the 3 following version of PyMySQL to the tests matrix:

```yaml
- '0.10.1'  # Released Sep 10, 2020 RHEL8-9 python3-PyMySQL
- '1.0.2'   # Released Jan 09, 2021 RHEL9 python3.11-PyMySQL
- '1.1.0'   # Released Jun 26, 2023 RHEL9 python3.12-PyMySQL
```

Also the current test matrix tests against Python 3.8, 3.9 and 3.10. But not for 3.11 and 3.12.

Instead of adding new python version for the whole matrix, I propose to take advantage of the `include` option for GitHub Action matrix: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude So something along the line:

```yaml
strategy:
  matrix:
    python:
      - '3.8'
      - '3.9'
      - '3.10'
    connector_version:
      - '0.7.11'   # Released Apr 05, 2017
      - '0.9.3'     # Released Dec 18, 2018
      - '2.0.1'
      - '2.0.3'
      - '2.1.1'
    include:
      - python: '3.6'  # RHEL8 uses 3.6 by default
        connector_version: '0.10.1'  # From RHEL package python3-PyMySQL
        connector_name: pymysql

      - python: '3.9'  # RHEL9 uses 3.9 by default
        connector_version: '0.10.1'  # From RHEL package python3-PyMySQL
        connector_name: pymysql

      - python: '3.11'
        connector_version: '1.0.2'  # From RHEL package python3.11-PyMySQL
        connector_name: pymysql

      - python: '3.12'
        connector_version: '1.1.0'  # From RHEL package python3.12-PyMySQL
        connector_name: pymysql
```

Or alternatively, we could add Python 3.11 to ansible-core 2.17 and 3.12 to devel ? This is a more generalized matrix and still keep the amount of tests to a reasonable count.

### LINKED ISSUE #625: GHA: add testing against MariaDB 10.11 LTS
##### SUMMARY

The collection should test every LTS version of MariaDB. Currently 10.11 is not tested

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

tests

##### ADDITIONAL INFORMATION

In order to limit the number of tests, I think when we add a new version of a component, the collection should use only latests version of the other components. Namely Python, ansible-core and the connectors. So the new matrix could be something like this:

```yaml
strategy:
  matrix:
    ansible:
      - stable-2.15
      - stable-2.16
      - stable-2.17
      - devel
    db_engine_name:
      - mysql
      - mariadb
    db_engine_version:
      - '5.7.40'
      - '8.0.31'
      - '10.4.27'
      - '10.5.18'
      - '10.6.11'
    python:
      - '3.8'
      - '3.9'
      - '3.10'
    connector_name:
      - pymysql
      - mysqlclient
    connector_version:
      - '0.7.11'  # Released Apr 05, 2017
      - '0.9.3'    # Released Dec 18, 2018
      - '1.0.2'
      - '2.0.1'
      - '2.0.3'
      - '2.1.1'
    include:
      - db_engine_version: 10.11.7
        db_engine_name: mariadb
        ansible: stable-2.17
        python: '3.10' 
        connector_version: '0.9.3'
        connector_name: pymysql

      - db_engine_version: 10.11.7
        db_engine_name: mariadb
        ansible: stable-2.17
        python: '3.10' 
        connector_version: '2.1.1'
        connector_name: mysqlclient

```

laurent-indermuehle:
## SUMMARY

fixes #624
fixes #625

### Removes tests for specific version of Python
why:
- Bug for a specific version of Python are rare in Integration tests
- This allow us to use ubuntu2204 instead of unbun2004. This is a limitation of ansible-test. The alternative would have been to use custom container, but we removes ours 3 weeks ago because it was too much maintenance.
- Ubuntu 22.04 can connect to MySQL 8.4. This is not the case for ubuntu 20.04.
- Reduce the number of tests
- It means we only use Python 3.10 (until ansible-test update its images).
- Fix CI for MySQL 8.0 that occured due to an update in ubuntu 20.04 last week (see bellow)

### Remove tests for
- MySQL 5.7.40 that is EoL
- MariaDB 10.4.27 that is EoL
- MariaDB 10.6.11, reason being that usually breaking changes introduced in the versions in between are also present in the latest version.
- PyMySQL 0.7.11 (it is very very old and we can't have too many tests)

### Add tests for
- RHEL8 context: pymysql:0.10.1, ansible:2.16
- MariaDB 10.11.8
- PyMySQL 1.1.1

### Update tests
- MySQL 8.0.31 -> 8.0.38
- MariaDB 10.5.18 -> 10.5.25

### Number of VM
- Before this PR; integration tests used 34 VMs.
- After this PR; integration tests use 36 VMs.

### Note for MySQL 8.4
Since MySQL 8.2, the mysql_native_plugin is absent. I modified the Makefile and GHA workflow to add the option to enable it back. Otherwise our tests fails to connect to the database.

Other tests are now failing due to deprecated option between 8.0 and 8.4 but I intend to let the tests red here and open a new PR to fix issues with MySQL 8.4 later.

### Note for MySQL 8.0
One week ago, our CI stopped working. This is because Ubuntu 20.04 got updated for security reasons and now mysql-client complains that 'cryptography' package is missing. The upgrade to ubuntu 22.04 fix that.

### Note for MySQL 9 and MariaDB 11
None of the latest db engine have a LTS release. So we will wait a little longer before testing those versions.

## ISSUE TYPE

- Tests

## COMPONENT NAME

- CI

### DIFF
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -275,9 +275,9 @@ jobs:
             repl1_conf='[mysqld]\\nserver-id=2\\nlog-bin=/var/lib/mysql/replica1-bin'
             repl2_conf='[mysqld]\\nserver-id=3\\nlog-bin=/var/lib/mysql/replica2-bin'
           fi
-          docker exec ${{ job.services.db_primary.id }} bash -c 'echo -e $prima_conf > /etc/mysql/conf.d/replication.cnf'
-          docker exec ${{ job.services.db_replica1.id }} bash -c 'echo -e $repl1_conf > /etc/mysql/conf.d/replication.cnf'
-          docker exec ${{ job.services.db_replica2.id }} bash -c 'echo -e $repl2_conf > /etc/mysql/conf.d/replication.cnf'
+          docker exec -e cnf=$prima_conf ${{ job.services.db_primary.id }} bash -c 'echo -e $cnf > /etc/mysql/conf.d/replication.cnf'
+          docker exec -e cnf=$repl1_conf ${{ job.services.db_replica1.id }} bash -c 'echo -e $cnf > /etc/mysql/conf.d/replication.cnf'
+          docker exec -e cnf=$repl2_conf ${{ job.services.db_replica2.id }} bash -c 'echo -e $cnf > /etc/mysql/conf.d/replication.cnf'
           docker restart -t 30 ${{ job.services.db_primary.id }}
           docker restart -t 30 ${{ job.services.db_replica1.id }}
           docker restart -t 30 ${{ job.services.db_replica2.id }}
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -275,9 +275,9 @@ jobs:
             repl1_conf='[mysqld]\\nserver-id=2\\nlog-bin=/var/lib/mysql/replica1-bin'
             repl2_conf='[mysqld]\\nserver-id=3\\nlog-bin=/var/lib/mysql/replica2-bin'
           fi
-          docker exec -e cnf=$prima_conf ${{ job.services.db_primary.id }} bash -c 'echo -e $cnf > /etc/mysql/conf.d/replication.cnf'
-          docker exec -e cnf=$repl1_conf ${{ job.services.db_replica1.id }} bash -c 'echo -e $cnf > /etc/mysql/conf.d/replication.cnf'
-          docker exec -e cnf=$repl2_conf ${{ job.services.db_replica2.id }} bash -c 'echo -e $cnf > /etc/mysql/conf.d/replication.cnf'
+          docker exec -e cnf=$prima_conf ${{ job.services.db_primary.id }} bash -c 'echo -e ${cnf//\\n/\n} > /etc/mysql/conf.d/replication.cnf'
+          docker exec -e cnf=$repl1_conf ${{ job.services.db_replica1.id }} bash -c 'echo -e ${cnf//\\n/\n} > /etc/mysql/conf.d/replication.cnf'
+          docker exec -e cnf=$repl2_conf ${{ job.services.db_replica2.id }} bash -c 'echo -e ${cnf//\\n/\n} > /etc/mysql/conf.d/replication.cnf'
           docker restart -t 30 ${{ job.services.db_primary.id }}
           docker restart -t 30 ${{ job.services.db_replica1.id }}
           docker restart -t 30 ${{ job.services.db_replica2.id }}
--- a/.github/workflows/ansible-test-plugins.yml
+++ b/.github/workflows/ansible-test-plugins.yml
@@ -271,6 +271,7 @@ jobs:
             echo -n "${{ matrix.ansible }}"
             > tests/integration/ansible
           testing-type: integration
+          integration-retry-on-error: false
 
   units:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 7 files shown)


.github/workflows/ansible-test-plugins.yml
```diff
@@ -53,16 +53,11 @@ jobs:
           - mariadb
         db_engine_version:
           - '8.0.38'
-          # - '8.4.1'  # Unable to install mysql-client on Ubuntu 20.04 default container used by ansible-test
+          - '8.4.1'
           - '10.4.34'
           - '10.5.25'
           - '10.6.18'
```
Andersson007:
```suggestion
```
should we remove these intermediate versions and use only the oldest supported and the newest versions we can?

laurent-indermuehle: These are LTS supported until 2025 and 2026.
But 10.4 is EoL, I missed that: https://endoflife.date/mariadb

Andersson007: Should we then remove 10.4 and 10.6.18 and keep 10.5?

Andersson007: the justification for keeping only the oldest supported and the latest is that usually breaking changes introduced in the versions in between are also present in the latest version

laurent-indermuehle: I like this idea! I'm doing it!

laurent-indermuehle: 34 vm before, 42 now. And we still test mysqlclient! I mark this as resolved. Thanks @Andersson007 for your valuable input!

.github/workflows/ansible-test-plugins.yml
```diff
@@ -50,142 +52,117 @@ jobs:
...
-          - 1.0.2
-          - 2.0.1
-          - 2.0.3
-          - 2.1.1
+          - '0.9.3'
+          - '1.0.2'
+          - '1.1.1'
+          - '2.0.1'
+          - '2.0.3'
+          - '2.1.1'
```
Andersson007, line 68:
```suggestion
          - '0.9.3'
          - '1.0.2'
          - '1.1.1'
          - '2.0.1'
          - '2.0.3'
          - '2.1.1'
```
can we also apply same oldest-supported-and-newest strategy here?

laurent-indermuehle: Best way to install PyMySQL is to use the distribution's package manager.
The version depends on the will of the APT/RPM maintainers:
- Ubuntu 22.04 : 1.0.2
- RHEL8 : 0.10.1
- RHEL9 : 1.1.1
- Fedora39 : 1.1.0-3 and 1.1.1-1

So we can't say: "This 1.0.2 bug was fixed in 1.1.1, please update", if their distribution doesn't provide such package.

We won't save that many VMs anyway. And PyMySQL is not updated that often: https://pypi.org/project/PyMySQL/#history

laurent-indermuehle:
Ready for review please.
The only failed test are for MySQL 8.4 and will be fixed later.

Andersson007 (approved):

PR #679, opened 2024-09-10
changelogs: categorize deprecations under deprecated_features

gotmax23:
##### SUMMARY
changelogs: categorize deprecations under deprecated_features

These should be put under deprecated_features so they show up properly in the generated changelog.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
changelogs

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/changelog.yaml
+++ b/changelogs/changelog.yaml
@@ -99,7 +99,7 @@ releases:
     release_date: '2022-04-26'
   3.10.0:
     changes:
-      breaking_changes:
+      deprecated_features:
       - collection - support of mysqlclient connector is deprecated - use PyMySQL
         connector instead! We will stop testing against it in collection version 4.0.0
         and remove the related code in 5.0.0 (https://github.com/ansible-collections/community.mysql/issues/654).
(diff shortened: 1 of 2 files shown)


changelogs/changelog.yaml
```diff
@@ -158,7 +158,7 @@ releases:
     release_date: '2024-08-22'
   3.10.1:
     changes:
-      breaking_changes:
+      deprecated_features:
       - mysql_user - the ``user`` alias of the ``name`` argument has been deprecated
```
felixfontein, line 162: Also 3.10.1 is a bugfix release, it should **not** have deprecations. See the semantic versioning spec: https://semver.org/

felixfontein (approved):

changelogs/changelog.yaml
laurent-indermuehle: @felixfontein , I think it was not a deprecation. It's a deprecation warning. The removal will happen in 5.0.
So the version was correct, right?

felixfontein: When you mark something as deprecated and add a waring, that's a deprecation, and according to https://semver.org/ must happen in a minor or major release, not in a bugfix release.

(The actual removal of the functionality must happen in a major release.)

Andersson007: cool, now we know it for future cases, thanks for the info!

Andersson007:
@gotmax23 thanks for fixing this!
@laurent-indermuehle @felixfontein thanks for reviewing!

PR #697, opened 2025-01-16
mysql_query: returns execution_time_ms list containing execution time per query

Andersson007:
##### SUMMARY
Was [requested](https://github.com/ansible-collections/community.postgresql/issues/787) for c.postgresql and I think it'd be nice to have this here too

### DIFF
--- a/changelogs/fragments/0-mysql_query-returns-exec-time-ms.yml
+++ b/changelogs/fragments/0-mysql_query-returns-exec-time-ms.yml
@@ -1,2 +1,2 @@
 minor_changes:
-- mysql_query - returns the `execution_time_ms` list containing execution time per query in milliseconds.
+- mysql_query - returns the ``execution_time_ms`` list containing execution time per query in milliseconds.
--- a/changelogs/fragments/0-mysql_query-returns-exec-time-ms.yml
+++ b/changelogs/fragments/0-mysql_query-returns-exec-time-ms.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- mysql_query - returns the `execution_time_ms` list containing execution time per query in milliseconds.
--- a/plugins/modules/mysql_query.py
+++ b/plugins/modules/mysql_query.py
@@ -62,7 +62,6 @@ author:
 - Andrew Klychkov (@Andersson007)
 extends_documentation_fragment:
 - community.mysql.mysql
-
 '''
 
 EXAMPLES = r'''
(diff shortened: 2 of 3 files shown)


Andersson007:
ready for review

markuman (approved):

laurent-indermuehle:
Nice addition @Andersson007. I think this is ready.

I'm surprised this feature is not described in DB API 2.0 (PEP 249). Otherwise PyMySQL could have take care of this.

changelogs/fragments/0-mysql_query-returns-exec-time-ms.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+- mysql_query - returns the `execution_time_ms` list containing execution time per query in milliseconds.
```
laurent-indermuehle:
```suggestion
- mysql_query - returns the ``execution_time_ms`` list containing execution time per query in milliseconds.
```

plugins/modules/mysql_query.py
```diff
@@ -139,6 +148,18 @@
...
 
+
+def execute_and_return_time(cursor, query, args):
+    # Measure query execution time in milliseconds
+    start_time = time.perf_counter()
+
+    cursor.execute(query, args)
+
+    # Calculate the execution time rounding it to 4 decimal places
+    exec_time_ms = round((time.perf_counter() - start_time) * 1000, 4)
```
laurent-indermuehle, line 159: Why *1000? Isn't time.perf_counter() already in ms?

laurent-indermuehle (commented):
I have just one question about line 159.
Since tests only checks if the returned value is bigger than 0, we never check what's returned.

plugins/modules/mysql_query.py
Andersson007: i found that it returns time in seconds:
```
The perf_counter() function always returns the float value of time in seconds.
```

Andersson007:
> I have just one question about line 159. Since tests only checks if the returned value is bigger than 0, we never check what's returned.

@laurent-indermuehle your suggestion has been committed, also answered the question

plugins/modules/mysql_query.py
Andersson007: in case of one query it returns
```
    "execution_time_ms": [
        6.248
    ],
```

in case of 3 queries:
```
    "execution_time_ms": [
        0.2793,
        0.1921,
        0.1112
    ],
```
Yeah it's floating so it's kinda tricky to tests against any hardcoded values

Andersson007: copied from local test run ^

laurent-indermuehle: I just get it. Line 153 is the doc for the method, not just line 154! So it make sense now. You can ignore my question. Thanks for the precision!

Andersson007: @laurent-indermuehle sure, if it's ok now, feel free to merge

Andersson007:
@markuman @laurent-indermuehle thanks for reviewing!
@optica-phoffmann thanks for requesting the feature!

Issue #701, opened 2025-02-27
Support user account locking


Keeper-of-the-Keys:
##### SUMMARY
Support `ACCOUNT LOCK | ACCOUNT UNLOCK` in `CREATE USER` and `ALTER USER`

This is to create "system" accounts for stored procedures, see also:
https://mariadb.com/kb/en/account-locking/
https://dev.mysql.com/doc/refman/8.4/en/account-locking.html

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
This would just extend the user module.

##### ADDITIONAL INFORMATION
I think this could either be an additional state ie.

```
state: (present | locked | absent)
```

Example

```
- name: Create user for stored procedure
  become: yes
  community.mysql.mysql_user:
    login_unix_socket: /var/run/mysqld/mysqld.sock
    name: 'procedure_definer'
    priv: '{{ db_name }}.*:SELECT'
    state: locked
```

Or an additional option:

```
- name: Create user for stored procedure
  become: yes
  community.mysql.mysql_user:
    login_unix_socket: /var/run/mysqld/mysqld.sock
    name: 'procedure_definer'
    priv: '{{ db_name }}.*:SELECT'
    locked: yes
    state: present
```

I looked a bit at `user.py` and it seems to me that the best way to implement it would be in **serial** as in first run the current user creation logic and then lock/unlock because the alternative would double the current list of potential create user queries from 10 to 20

If I have time and there is a positive response to this issue I'll try to create a PR.

Andersson007:
@Keeper-of-the-Keys hello, thanks for the feature request!
We have a limited number of hands here, adding new features when we have time.
In the meantime, if anybody wants to implement this, there's the [quick-start guide](https://docs.ansible.com/ansible/devel/community/create_pr_quick_start.html) that can help to take off.
~~I'll put the help wanted label, so any volunteers to dig deeper are welcome.~~

UPDATE: ah, i see the last line now, @laurent-indermuehle and others, what do you think?

laurent-indermuehle:
Hi @Keeper-of-the-Keys, thanks for the idea and most of all for your offer to contribute.
Just to give me an idea, will it make a difference between `state: absent` and `state: locked`?
I know that some things (views, stores procedures, ...) breaks if the definer account is absent. But how MariaDB/MySQL react to locked accounts?
From an Ansible Perspective, there is none. To unlock/recreate the account, in both case you use `state: present`.
[...]

PR #702, opened 2025-03-03
User locking

Keeper-of-the-Keys:
##### SUMMARY
Adds support for user locking as discussed in #701 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
user

### DIFF
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -61,9 +61,8 @@ def user_is_locked(cursor, user, host, host_all):
     # Unless I am very much mistaken there should only be 1 answer to this query ever.
     result = cursor.fetchone()
 
-    for res in result.values():
-        if res.endswith('ACCOUNT LOCK'):
-            return True
+    if result[0].endswith('ACCOUNT LOCK'):
+        return True
 
     return False
 
@@ -176,7 +175,7 @@ def get_existing_authentication(cursor, user, host=None):
 def user_add(cursor, user, host, host_all, password, encrypted,
              plugin, plugin_hash_string, plugin_auth_string, salt, new_priv,
              attributes, tls_requires, reuse_existing_password, module,
-             password_expire, password_expire_interval):
+             password_expire, password_expire_interval, locked=False):
     # If attributes are set, perform a sanity check to ensure server supports user attributes before creating user
     if attributes and not get_attribute_support(cursor):
         module.fail_json(msg="user attributes were specified but the server does not support user attributes")
@@ -266,8 +265,8 @@ def user_add(cursor, user, host, host_all, password, encrypted,
         cursor.execute("ALTER USER %s@%s ATTRIBUTE %s", (user, host, json.dumps(attributes)))
         final_attributes = attributes_get(cursor, user, host)
 
-#    if locked:
-#        cursor.execute("ALTER USER %s@%s ACCOUNT LOCK", (user, host))
+    if locked:
+        cursor.execute("ALTER USER %s@%s ACCOUNT LOCK", (user, host))
 
     return {'changed': True, 'password_changed': not used_existing_password, 'attributes': final_attributes}
 
@@ -283,7 +282,7 @@ def is_hash(password):
 def user_mod(cursor, user, host, host_all, password, encrypted,
              plugin, plugin_hash_string, plugin_auth_string, salt, new_priv,
              append_privs, subtract_privs, attributes, tls_requires, module,
-             password_expire, password_expire_interval, role=False, maria_role=False):
+             password_expire, password_expire_interval, locked=False, role=False, maria_role=False):
     changed = False
     msg = "User unchanged"
     grant_option = False
@@ -555,6 +554,15 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
             if attribute_support:
                 final_attributes = attributes_get(cursor, user, host)
 
+        if user_is_locked(cursor, user, host, False) != locked:
+            if locked:
+                cursor.execute("ALTER USER %s@%s ACCOUNT LOCK", (user, host))
+                msg = 'User locked'
+            else:
+                cursor.execute("ALTER USER %s@%s ACCOUNT UNLOCK", (user, host))
+                msg = 'User unlocked'
+            changed = True
+
         if role:
             continue
 
@@ -577,13 +585,6 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
 
                 cursor.execute(*query_with_args)
             changed = True
-
-#        if user_is_locked(cursor, user, host, False) != locked:
-#            if locked:
-#                cursor.execute("ALTER USER %s@%s ACCOUNT LOCK", (user, host))
-#            else:
-#                cursor.execute("ALTER USER %s@%s ACCOUNT UNLOCK", (user, host))
-#            changed = True
 
     return {'changed': changed, 'msg': msg, 'password_changed': password_changed, 'attributes': final_attributes}
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -566,6 +566,12 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
                 else:
                     cursor.execute("ALTER USER %s@%s ACCOUNT UNLOCK", (user, host))
                     msg = 'User unlocked'
+            else:
+                if locked:
+                    msg = 'User will be locked'
+                else:
+                    msg = 'User will be unlocked'
+
             changed = True
 
         if role:
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -52,16 +52,14 @@ def user_exists(cursor, user, host, host_all):
     return count[0] > 0
 
 
-def user_is_locked(cursor, user, host, host_all):
-    if host_all:
-        cursor.execute("SHOW CREATE USER %s", (user,))
-    else:
-        cursor.execute("SHOW CREATE USER %s@%s", (user, host))
+def user_is_locked(cursor, user, host):
+    cursor.execute("SHOW CREATE USER %s@%s", (user, host))
 
-    # Unless I am very much mistaken there should only be 1 answer to this query ever.
+    # Per discussions on irc:libera.chat:#maria the query may return up to 2 rows but "ACCOUNT LOCK" should always be in the first row.
     result = cursor.fetchone()
 
-    if result[0].endswith('ACCOUNT LOCK'):
+    # ACCOUNT LOCK does not have to be the last option in the CREATE USER query.
+    if result[0].find('ACCOUNT LOCK') > 0:
         return True
 
     return False
@@ -554,7 +552,7 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
             if attribute_support:
                 final_attributes = attributes_get(cursor, user, host)
 
-        if user_is_locked(cursor, user, host, False) != locked:
+        if user_is_locked(cursor, user, host) != locked:
             if locked:
                 cursor.execute("ALTER USER %s@%s ACCOUNT LOCK", (user, host))
                 msg = 'User locked'
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -58,12 +58,15 @@ def user_is_locked(cursor, user, host):
     # Per discussions on irc:libera.chat:#maria the query may return up to 2 rows but "ACCOUNT LOCK" should always be in the first row.
     result = cursor.fetchone()
 
-    from pprint import pprint
-    pprint(result)
-
     # ACCOUNT LOCK does not have to be the last option in the CREATE USER query.
-    if result[0] and result[0].find('ACCOUNT LOCK') > 0:
-        return True
+    # Need to handle both DictCursor and non-DictCursor
+    if type(result) == type([]):
+        if result[0] and result[0].find('ACCOUNT LOCK') > 0:
+            return True
+    elif type(result) == type({}):
+        for res in result.values():
+            if res.find('ACCOUNT LOCK') > 0:
+                return True
 
     return False
 
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -58,8 +58,11 @@ def user_is_locked(cursor, user, host):
     # Per discussions on irc:libera.chat:#maria the query may return up to 2 rows but "ACCOUNT LOCK" should always be in the first row.
     result = cursor.fetchone()
 
+    from pprint import pprint
+    pprint(result)
+
     # ACCOUNT LOCK does not have to be the last option in the CREATE USER query.
-    if result[0].find('ACCOUNT LOCK') > 0:
+    if result[0] and result[0].find('ACCOUNT LOCK') > 0:
         return True
 
     return False
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -553,12 +553,13 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
                 final_attributes = attributes_get(cursor, user, host)
 
         if user_is_locked(cursor, user, host) != locked:
-            if locked:
-                cursor.execute("ALTER USER %s@%s ACCOUNT LOCK", (user, host))
-                msg = 'User locked'
-            else:
-                cursor.execute("ALTER USER %s@%s ACCOUNT UNLOCK", (user, host))
-                msg = 'User unlocked'
+            if not module.check_mode:
+                if locked:
+                    cursor.execute("ALTER USER %s@%s ACCOUNT LOCK", (user, host))
+                    msg = 'User locked'
+                else:
+                    cursor.execute("ALTER USER %s@%s ACCOUNT UNLOCK", (user, host))
+                    msg = 'User unlocked'
             changed = True
 
         if role:
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -60,10 +60,10 @@ def user_is_locked(cursor, user, host):
 
     # ACCOUNT LOCK does not have to be the last option in the CREATE USER query.
     # Need to handle both DictCursor and non-DictCursor
-    if type(result) == type([]):
+    if type(result) is list:
         if result[0] and result[0].find('ACCOUNT LOCK') > 0:
             return True
-    elif type(result) == type({}):
+    elif type(result) is dict:
         for res in result.values():
             if res.find('ACCOUNT LOCK') > 0:
                 return True
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -61,13 +61,14 @@ def user_is_locked(cursor, user, host):
     # ACCOUNT LOCK does not have to be the last option in the CREATE USER query.
     # Need to handle both DictCursor and non-DictCursor
     if isinstance(result, list):
-        if result[0] and result[0].find('ACCOUNT LOCK') > 0:
+        if result[0].find('ACCOUNT LOCK') > 0:
             return True
     elif isinstance(result, dict):
         for res in result.values():
             if res.find('ACCOUNT LOCK') > 0:
                 return True
-
+    from pprint import pprint
+    pprint(result)
     return False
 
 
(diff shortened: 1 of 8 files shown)


Keeper-of-the-Keys:
I suspect that the `host_all` option in `user_is_locked` can be dropped due to where in the code path the function is called.

laurent-indermuehle:
About the host_all. I would say, it depends. Can we lock roles too? What if multiple accounts with the same name are present?
I think the simpler way to know for sure what we are doing is by adding integration tests for all possible scenarii. This way we will see if the accounts locking works the same between MySQL, MariaDB and their differents supported version.

Keeper-of-the-Keys:
@laurent-indermuehle can you point me to where I would need to write the integration test (and an example from which I can extend to these functions)?

At the moment `user_is_locked` is called from **inside** the loop on hosts ([`for host in hosts:`](https://github.com/ansible-collections/community.mysql/blob/e9845b0a1caba4344aab9e957865ac74ab17fc7f/plugins/module_utils/user.py#L282)) so unless there is some reason that you can forsee to call the function outside the loop I think I can drop that code (I initially modelled the function on `user_exists` but based on the actual usage I don't think the `hosts_all` arg and conditional is needed).

As for roles - I don't know I did not go into that much depth.

laurent-indermuehle:
Ok let's forget roles for now. We can focus on normal accounts.

I would create a file called: `tests/integrations/targets/test_mysql_user/tasks/test_user_locking.yml`
Then you must add it to `tests/integrations/targets/test_mysql_user/tasks/main.yml` at the bottom.

To get a good example, I would look at `tests/integrations/targets/test_mysql_user/tasks/test_column_case_sensitive.yml` as it is one of the latest we wrote. It uses `failed_when` instead of `assert` for better log outputs.
But you can get examples in any other tasks file in that directory.

Then, to either run tests locally (the fastest) or here in GitHub Action (the easiest), you can use our guides: https://github.com/ansible-collections/community.mysql/blob/main/TESTING.md

Keeper-of-the-Keys:
@laurent-indermuehle 
If I want to add the `locked` attribute to the doc to fix the sanity checks what version should I put in for `version_added`?

laurent-indermuehle:
I'm preparing a PR myself. I think next version will be 3.13

Keeper-of-the-Keys:
Should I also add an example or just attribute documentation?

plugins/modules/mysql_user.py
```diff
@@ -470,6 +478,7 @@ def main():
         column_case_sensitive=dict(type='bool', default=None),  # TODO 4.0.0 add default=True
         password_expire=dict(type='str', choices=['now', 'never', 'default', 'interval'], no_log=True),
         password_expire_interval=dict(type='int', required_if=[('password_expire', 'interval', True)], no_log=True),
+        locked=dict(type='bool', default='no'),
```
```suggestion
        locked=dict(type='bool', default='false'),
```
it's officially recommended

plugins/modules/mysql_user.py
```diff
@@ -189,6 +189,14 @@
         fields names in privileges.
     type: bool
     version_added: '3.8.0'
+
+  locked:
+    description:
+      - Lock account to prevent connections using it, this is primarily used for creating a user that will act as a DEFINER on stored procedures.
+      - The default is C(false)
```
```suggestion
    default: false
```

tests/integration/targets/test_mysql_user/tasks/test_user_locking.yml
```diff
@@ -0,0 +1,140 @@
...
+
+    # ========================== Tests ========================================
+
+    - name: Mysql_user Lock user | Create test user locked
+      community.mysql.mysql_user:
+        <<: *mysql_params
+        name: mysql_locked_user
+        host: '%'
+        password: 'msandbox'
+        locked: yes
```
Andersson007:
```suggestion
        locked: yes
```
true/false is a recommended way since some time ago, officially. it's not critical here, but in modules docs and files it must be `true/false`

Andersson007 (commented):
@Keeper-of-the-Keys thanks for the PR!
Could you also please:
- Add a changelog [fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)
- Add test cases with check_mode: true and check if the changes haven't been made

Andersson007:
> Should I also add an example or just attribute documentation?

Yes, please add a separate dedicated example

Keeper-of-the-Keys:
Hey I just realized why the integration tests were failing - in `user_is_locked` I was assuming that `ACCOUNT LOCK` was the last option in the query.

In the last commit I changed `.endswith('ACCOUNT LOCK')` to `.find('ACCOUNT LOCK') >0`, I also removed the `host_all` flag from the function since that was not really in use.

If this is desired and one of you has time to give me some guidance I don't mind also trying to fix the codecov complaint but it is not obvious to me where that lives.

Keeper-of-the-Keys:
@laurent-indermuehle @Andersson007 Some of the integration tests seem to be failing on unrelated errors, at least based on me clicking some at random they seem to fail on `test_mysql_variables`) is there an issue with that test?

Earlier I had errors where users were not unlocked but now it seems like issues unrelated to my changes.

laurent-indermuehle:
@Keeper-of-the-Keys 
Many errors are ignored (the blue-green text ...ignored). Don't fall into the trap of reading those (this is why I prefer failed_when over assert!).

I see those 3 errors is the latest integration tests:
```
TASK [test_mysql_role : Subtract privileges that are not in the current privileges, which 
should be a no-op] ***
fatal: [testhost]: FAILED! => {"changed": false, "msg": "(1133, \"Can't find any 
matching row in the user table\")"}
```

```
TASK [test_mysql_user : Mysql_user Lock user | create locked | Assert that test user is locked] ***
fatal: [testhost]: FAILED! => {"changed": false, "failed_when_result": "The conditional check 
'locked_user_creation.query_result[0][0] is not search('ACCOUNT LOCK')' failed. The error 
was: error while evaluating conditional (locked_user_creation.query_result[0][0] is not 
search('ACCOUNT LOCK')): 'dict object' has no attribute 'query_result'", "msg": "Cannot 
execute SQL 'SHOW CREATE USER 'mysql_locked_user'@'%'' args [None]: (1133, \"Can't
find any matching row in the user table\")"}
``` 

```
TASK [test_mysql_variables : Get server version] *******************************
An exception occurred during task execution. To see the full traceback, use -vvv. The 
error was: pymysql.err.OperationalError: (1141, "There is no such grant defined for 
user 'role2' on host '%'")
```

I only read (Ⓐstable-2.17, DB: mariadb 10.11.8, connector: pymysql 1.1.1), maybe for MySQL the error related to roles are different/absent?

Keeper-of-the-Keys:
@laurent-indermuehle I believe I resolved my tests, I am puzzled by the role error you show, `user_mod()` does not do anything with roles see [here](https://github.com/ansible-collections/community.mysql/blob/e9845b0a1caba4344aab9e957865ac74ab17fc7f/plugins/module_utils/user.py#L539).

As an aside I would suggest that the calls to `user_mod` that have long sequences of `None, None` in the calling arguments should use named args for everything, I can add that to this PR:
```diff
--- a/plugins/modules/mysql_role.py
+++ b/plugins/modules/mysql_role.py
@@ -930,11 +930,12 @@ class Role():
                                                       set_default_role_all=set_default_role_all)
 
         if privs:
-            result = user_mod(self.cursor, self.name, self.host,
-                              None, None, None, None, None, None, None,
-                              privs, append_privs, subtract_privs, None, None,
-                              self.module, None, None, role=True,
-                              maria_role=self.is_mariadb)
+            result = user_mod(cursor=self.cursor, user=self.name, host=self.host,
+                              host_all=None, password=None, encrypted=None, plugin=None,
+                              plugin_auth_string=None, plugin_hash_string=None, salt=None,
+                              new_priv=privs, append_privs=append_privs, subtract_privs=subtract_privs,
+                              attributes=None, tls_requires=None, module=self.module, password_expire=None,
+                              password_expire_interval=None, role=True, maria_role=self.is_mariadb)
             changed = result['changed']
 
         if admin:
--- a/plugins/modules/mysql_user.py
+++ b/plugins/modules/mysql_user.py
@@ -594,13 +594,15 @@ def main():
                     result = user_mod(cursor, user, host, host_all, password, encrypted,
                                       plugin, plugin_hash_string, plugin_auth_string, salt,
                                       priv, append_privs, subtract_privs, attributes, tls_requires, module,
-                                      password_expire, password_expire_interval, locked)
+                                      password_expire, password_expire_interval, locked=locked)
 
                 else:
-                    result = user_mod(cursor, user, host, host_all, None, encrypted,
-                                      None, None, None, None,
-                                      priv, append_privs, subtract_privs, attributes, tls_requires, module,
-                                      password_expire, password_expire_interval, locked)
+                    result = user_mod(cursor=cursor, user=user, host=host, host_all=host_all, password=None,
+                                      encrypted=encrypted, plugin=None, plugin_hash_string=None, plugin_auth_string=None,
+                                      salt=None, new_priv=priv, append_privs=append_privs, subtract_privs=subtract_privs,
+                                      attributes=attributes, tls_requires=tls_requires, module=module,
+                                      password_expire=password_expire, password_expire_interval=password_expire_interval,
+                                      locked=locked)

```

laurent-indermuehle:
@Keeper-of-the-Keys we should avoid mixing different things in the same PR. 

Regarding the error about roles. I'm getting the exact same error on my PR and locally. It's only happening with MariaDB: pymysql.err.OperationalError: (1141, "There is no such grant defined for user 'PUBLIC' on host '%'"). Maybe Docker did change something in there library/MariaDB containers !?! I'll continue to investigate.

You PR start to look good! You are doing a great job so far! I'll try to review your changes today.

plugins/modules/mysql_user.py
```diff
@@ -189,6 +189,14 @@
         fields names in privileges.
     type: bool
     version_added: '3.8.0'
+
+  locked:
+    description:
+      - Lock account to prevent connections using it, this is primarily used for creating a user that will act as a DEFINER on stored procedures.
```
```suggestion
      - Lock account to prevent connections using it.
      - This is primarily used for creating a user that will act as a DEFINER on stored procedures.
```

plugins/module_utils/user.py
```diff
@@ -536,6 +552,15 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
             if attribute_support:
                 final_attributes = attributes_get(cursor, user, host)
 
+        if user_is_locked(cursor, user, host) != locked:
```
```suggestion
        if user_is_locked(cursor, user, host) != locked:
```
this function should handle `module.check_mode`, see a bit below how it can be implemented

Andersson007 (commented):
@Keeper-of-the-Keys thanks! the following is still relevant. See my comment below about handling `module.check_mode` 
- Add test cases with check_mode: true and check if the changes haven't been made

plugins/module_utils/user.py
Keeper-of-the-Keys: Sorry missed that before, pushed a change.

Keeper-of-the-Keys:
@Andersson007 I added a test case for locking/unlocking should there also be a testcase where no action is needed? (ie. `locked: true` on an already locked account?

Which I assume would fail if changed?

laurent-indermuehle:
Yes please @Keeper-of-the-Keys. Tests should cover every possible scenarios.
Yes, having a test that checks for idempotence is a common usage!

Keeper-of-the-Keys:
@laurent-indermuehle done

laurent-indermuehle:
Great.
Now I think we must do what you suggested earlier about the `mysql_role` module. Because you added a new argument (locked) in the middle of the argument chain, you need to adapt the calls to user_mod in plugins/modules/mysql_role.py line 933 @Keeper-of-the-Keys.

Keeper-of-the-Keys:
@laurent-indermuehle sure I can do that, but unless I am very much mistaken it is not needed because the role arguments are already named, my suggestion to make everything named was mainly because I had to copy the function declaration and check all the positional arguments to make sure that I was correct.

(Also once we switch to named all the `None`s that are `default=None` can be dropped)

Keeper-of-the-Keys:
@laurent-indermuehle @Andersson007 Is there anything further that I can do to help push this over the edge?

laurent-indermuehle:
@Keeper-of-the-Keys sorry but you haven't finished, there is errors regarding roles to fix. Maybe you can't see the CI results? For me they are displayed just above the field "Add a comment" in this conversation and starts with "Plugins CI / Integration(...)".
Let me know if you can't see them.

Keeper-of-the-Keys:
@laurent-indermuehle sorry I was under the impression that that issue was happening regardless of this PR which is why I ignored the CI, I made the change as requested but as I pointed out `user_mod` for a **role** does nothing and `mysql_role` does not call `user_add` so I am at a loss to explain why this PR would be the cause of that failure but I'll look again.

Keeper-of-the-Keys:
I think I figured it out, just pushed the fix, `user_is_locked` was running always and would fail for roles.

Keeper-of-the-Keys:
@laurent-indermuehle As far as I can tell the issue with `test_mysql_role` is resolved.

As an aside - any idea why search on the CI output does not work? I need to scroll through the output to find the failed tests every time and when I try to search for a string I know is there (like `test_mysql_role` or `no-op`) it say 0 results unless that bit of output is showing, which was a contributing factor to me having more difficulty interacting with the CI tests.

(I'm using Firefox)

laurent-indermuehle:
@Keeper-of-the-Keys great investigation! And Thank you for your time!

I feel your pain about the output. I have difficulties too (using Vivaldi). I think you have to let it finish loading before you can search. Searching too soon make it stop loading.
Alternatively, you can download the output, or "view raw log". You lose the colors, but at least you're able the search !

laurent-indermuehle:
I've found one issue:
```yaml
TASK [test_mysql_user : Mysql_user Lock user | create locked | Unlock test user
 check_mode: true] ***
fatal: [testhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters 
for (community.mysql.mysql_user) module: check_mode. Supported parameters
 include: append_privs, attributes, ca_cert, check_hostname, check_implicit_admin,
 client_cert, client_key, column_case_sensitive, config_file, connect_timeout, 
encrypted, force_context, host, host_all, locked, login_host, login_password, 
login_port, login_unix_socket, login_user, name, password, password_expire, 
password_expire_interval, plugin, plugin_auth_string, plugin_hash_string, priv, 
resource_limits, salt, session_vars, sql_log_bin, state, subtract_privs, tls_requires, 
update_password (ssl_ca, ssl_cert, ssl_key, user)."}
```
I'm off now, I'll continue tomorrow. Feel free to ping me in case you need my help.

Keeper-of-the-Keys:
There is an error now in the the `mysql_info` module that involves a user entry that has the following hash: 

> "$A$005$THISISACOMBINATIONOFINVALIDSALTANDPASSWORDTHATMUSTNEVERBRBEUSED"

I have tried to find that in the setup of different test cases to no avail, is it part of the schema that is imported for test cases?

It looks like the user is setup locked and since lock/unlock logic was now added and the default is to unlock it gets unlocked failing the test, I guess the info module also needs a field of whether the account is locked.

Keeper-of-the-Keys:
@laurent-indermuehle @Andersson007 This is an interesting issue - 

The cursor context in mysql_info and mysql_user is different, thus in mysql_user I can refer to `result[0]` but in `mysql_info` it is a dict, I think this is due to the mysql_info cursor being a `DictCursor` where the mysql_user one is not.

While reimplementing the user_is_locked function for a DictCursor is possible I do place a serious question mark on having cursors behave differently in different modules but still importing functions across modules.

Keeper-of-the-Keys:
Not sure why the idempotence check failed, need to go now hope to continue later today.

Keeper-of-the-Keys:
@laurent-indermuehle @Andersson007 I think this is now ready...

Andersson007 (approved):
@Keeper-of-the-Keys thanks for the persistence!
Do you feel committed to fix related things if there are reported?

Keeper-of-the-Keys:
@Andersson007 Obviously I can't predict the future but where I can I maintain my code.

Andersson007:
> @Andersson007 Obviously I can't predict the future but where I can I maintain my code.

sure, if there are any issues, they are usually reported quickly in cases of such key modules:)

@laurent-indermuehle any other concerns/suggestions?

Keeper-of-the-Keys:
There is one potential issue, but it's really just a communication of changes to users -
Because it defaults to unlocked (which is current behavior) users who are currently manually locking accounts (with `mysql_query`) would now find that is `mysql_user` runs on that user after it it would unlock the account.

laurent-indermuehle (approved):

laurent-indermuehle:
Good point @Keeper-of-the-Keys. We should add this information (unlocking of existing accounts) in the release summary!
Does that means we can release this as 3.13 or should we wait for 4.0 @Andersson007?

I reviewed the code, this is really great work @Keeper-of-the-Keys! Thanks a lot for your contribution.

I feel like we should decide once and for all between Dict cursor and the other mode (don't recall the name) and align all the module to use the same one. I recall having lost time on this very trap... Sorry that you had to endure the same thing.

Andersson007:
@Keeper-of-the-Keys @laurent-indermuehle great point.
It's a breaking change then. I suggest not using False by default. We should not use any defaults.
In this case the module should do not run the related code. Only if it's explicitly specified as true/false it should run.

@Keeper-of-the-Keys could you incorporate the changes please?

Keeper-of-the-Keys:
@Andersson007 so `default(omit)`, can you refer me to an example of that, because that is no longer a simple `bool`?

Andersson007:
> @Andersson007 so `default(omit)`, can you refer me to an example of that, because that is no longer a simple `bool`?

@Keeper-of-the-Keys if you just remove `default: false` from the docs and everywhere, it'll be assigned with `None` IIRC. So you could use `if locked is not None: then run the code`.

Keeper-of-the-Keys:
Interesting, I would have expected that to fail due to an expected arg not being supplied.

I guess in the call to `user_add` and `user_mod` I would need to specify `None`.

Keeper-of-the-Keys:
@Andersson007 correction - the behavior would only need to be modified for `user_mod`, for `user_add` `default=False` is the same as the old behavior, how would I document that?

```
user_add - new user, False -> don't lock
user_mod - existing user, None -> don't touch lock (False- > unlock).
```

Andersson007:
> @Andersson007 correction - the behavior would only need to be modified for `user_mod`, for `user_add` `default=False` is the same as the old behavior, how would I document that?
> 
> ```
> user_add - new user, False -> don't lock
> user_mod - existing user, None -> don't touch lock (False- > unlock).
> ```

I would suggest:
```
If not specified, it does not change the lock state.
````
or something like that.

Keeper-of-the-Keys:
@Andersson007 I think it is done.

I added a test to check that a previously created locked account doesn't get unlocked if locked is left unspecified.

Andersson007 (approved):

Andersson007:
@Keeper-of-the-Keys thanks!
@laurent-indermuehle do you have anything else on your mind? If not, please merge

laurent-indermuehle:
Hi @Andersson007, nothing in my mind. @Keeper-of-the-Keys and you have done a good job!
I was contemplating the idea of changing the default to false in 4.0.0, but the test with None seems future proof.

Merging now. This will be included in 3.13 or 3.12.1, not sure yet. Thanks again for your contribution @Keeper-of-the-Keys.

Keeper-of-the-Keys:
@laurent-indermuehle note that I documented the feature as introduced in 3.13.0 so if you end up doing 3.12.1 that would need to be changed.

Andersson007:
> Hi @Andersson007, nothing in my mind. @Keeper-of-the-Keys and you have done a good job! I was contemplating the idea of changing the default to false in 4.0.0, but the test with None seems future proof.
> 
> Merging now. This will be included in 3.13 or 3.12.1, not sure yet. Thanks again for your contribution @Keeper-of-the-Keys.

@laurent-indermuehle i was also contemplating on the warning + default false in 4.0.0, but then i was thinking about all of those nasty warnings our users would get.. 
It should be 3.13 as the change is minor. Any plans to release it? I could do it on Friday if i don't forget

Andersson007:
@Keeper-of-the-Keys yes, it's gonna be 3.13 ^
Thanks for the contribution!

Andersson007:
@Keeper-of-the-Keys we also hang out on [Matrix](https://docs.ansible.com/ansible/devel/community/communication.html#real-time-chat) in the `#mysql:ansible.com` channel. If you're not already there. Just a fyi. You could also join `#postgresql:ansible.com` if you use the c.postgresql collection:)

PR #703, opened 2025-03-05
Add support for MariaDB 11.4

laurent-indermuehle:
##### SUMMARY

I'm adding MariaDB 11.4 to our test suite (and removing 10.5 as we only keep 2 versions now).
All tests pass, but that's because our Ubuntu containers have mysql-client installed by our test setup.
If you try to use mysql_db against a target that only has MariaDB 11+ server and client installed, you'll get 'mysql command not found'.
Fortunately, this is the only module that uses the binary directly. All other modules communicate through PyMySQL, so everything keeps working there, even with old versions of PyMySQL!

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

mysql_db

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -387,12 +387,13 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
             dump_extra_args=None, unsafe_password=False, restrict_config_file=False,
             check_implicit_admin=False, pipefail=False):
 
-    if server_implementation == 'mysql':
-        cmd = module.get_bin_path('mysqldump', True)
-    elif server_implementation == 'mariadb' and LooseVersion(server_version) >= LooseVersion("10.4.6"):
-        cmd = module.get_bin_path('mariadb-dump', True)
-    else:
-        return module.fail_json(msg="Unknown server implementation %s" % server_implementation)
+    cmd_str = 'mysqldump'
+    if server_implementation == 'mariadb' and LooseVersion(server_version) >= LooseVersion("10.4.6"):
+        cmd_str = 'mariadb-dump'
+    try:
+        cmd = module.get_bin_path(cmd_str, True)
+    except Exception as e:
+        return 1, "", "Error determining dump command: %s" % str(e)
 
     # If defined, mysqldump demands --defaults-extra-file be the first option
     if config_file:
@@ -490,12 +491,13 @@ def db_import(module, host, user, password, db_name, target, all_databases, port
     if not os.path.exists(target):
         return module.fail_json(msg="target %s does not exist on the host" % target)
 
-    if server_implementation == 'mysql':
-        cmd = [module.get_bin_path('mysql', True)]
-    elif server_implementation == 'mariadb' and LooseVersion(server_version) >= LooseVersion("10.4.6"):
-        cmd = [module.get_bin_path('mariadb', True)]
-    else:
-        return module.fail_json(msg="Unknown server implementation %s" % server_implementation)
+    cmd_str = 'mysql'
+    if server_implementation == 'mariadb' and LooseVersion(server_version) >= LooseVersion("10.4.6"):
+        cmd_str = 'mariadb'
+    try:
+        cmd = module.get_bin_path(cmd_str, True)
+    except Exception as e:
+        return 1, "", "Error determining mysql/mariadb command: %s" % str(e)
 
     # --defaults-file must go first, or errors out
     if config_file:
--- a/plugins/modules/mysql_db.py
+++ b/plugins/modules/mysql_db.py
@@ -391,7 +391,7 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
     if server_implementation == 'mariadb' and LooseVersion(server_version) >= LooseVersion("10.4.6"):
         cmd_str = 'mariadb-dump'
     try:
-        cmd = module.get_bin_path(cmd_str, True)
+        cmd = [module.get_bin_path(cmd_str, True)]
     except Exception as e:
         return 1, "", "Error determining dump command: %s" % str(e)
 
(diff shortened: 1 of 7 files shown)


laurent-indermuehle:
@Andersson007 any idea why integrations tests didn't run? Or how to force them?

laurent-indermuehle:
Note to self:

```yaml
    - name: Get MySQL settings
      community.mysql.mysql_info:
        filter:
          - users_info
```
Produce: pymysql.err.OperationalError: (1141, "There is no such grant defined for user 'PUBLIC' on host '%'")

- Why our tests didn't catch this?
- How to fix that?

Andersson007:
@laurent-indermuehle i see they run on schedule and in another PR. Can it be related to the Makefile changes?

laurent-indermuehle:
@Andersson007 I don't see the Makefile doing anything related to GH Actions.
I modified a file in the path of the to workflow event "push" but it didn't triggered the workflow O_o

EDIT found the issue: The workflow is not valid. .github/workflows/ansible-test-plugins.yml (Line: 180, Col: 18): Unrecognized named-value: 'env'. Located at position 1 within expression: env.HEALTH_CMD .

noonedeadpunk:
Just hit the same issue and was about to start writing a patch for it :)

plugins/modules/mysql_db.py
```diff
@@ -386,7 +386,14 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
             encoding=None, force=False, master_data=0, skip_lock_tables=False,
             dump_extra_args=None, unsafe_password=False, restrict_config_file=False,
             check_implicit_admin=False, pipefail=False):
-    cmd = module.get_bin_path('mysqldump', True)
+
+    if server_implementation == 'mysql':
+      cmd = module.get_bin_path('mysqldump', True)
+    elif server_implementation == 'mariadb':
```
noonedeadpunk: I would also check that server_version >= 11.0 as otherwise it might break MariaDB 10.1 or smth like that.

plugins/modules/mysql_db.py
```diff
@@ -476,13 +483,20 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
...
+              encoding=None, force=False,
               use_shell=False, unsafe_password=False, restrict_config_file=False,
               check_implicit_admin=False):
     if not os.path.exists(target):
         return module.fail_json(msg="target %s does not exist on the host" % target)
 
-    cmd = [module.get_bin_path('mysql', True)]
+    if server_implementation == 'mysql':
+      cmd = [module.get_bin_path('mysql', True)]
+    elif server_implementation == 'mariadb':
```
noonedeadpunk: ditto

plugins/modules/mysql_db.py
laurent-indermuehle: Thanks @noonedeadpunk for the suggestion. But didn't mariadb's binaries been present since day one?

laurent-indermuehle: You were right @noonedeadpunk! They were introduced in 10.4.6: https://mariadb.com/kb/en/mariadb-1046-release-notes/
I'm implementing the version check right now!

plugins/modules/mysql_db.py
```diff
@@ -386,7 +386,14 @@ def db_dump(module, host, user, password, db_name, target, all_databases, port,
             encoding=None, force=False, master_data=0, skip_lock_tables=False,
             dump_extra_args=None, unsafe_password=False, restrict_config_file=False,
             check_implicit_admin=False, pipefail=False):
-    cmd = module.get_bin_path('mysqldump', True)
+
+    if server_implementation == 'mysql':
```
noonedeadpunk: I think you'd need to adjust logic here? Ie:
```
    if server_implementation == 'mysql' or (server_implementation == 'mariadb' and LooseVersion(server_version) < LooseVersion("10.4.6")):
        cmd = module.get_bin_path('mysqldump', True)
    elif server_implementation == 'mariadb' and LooseVersion(server_version) >= LooseVersion("10.4.6"):
        cmd = module.get_bin_path('mariadb-dump', True)
```

As current will still fall under `Unknown server implementation`?

Or, detect prior version as `mysql` in here and don't bother with complex logic: https://github.com/ansible-collections/community.mysql/blob/e9845b0a1caba4344aab9e957865ac74ab17fc7f/plugins/module_utils/mysql.py#L217-L221

laurent-indermuehle:
Ready for review!

The issue I was having yesterday is impossible to reproduce (Our collection will fail if someone tries to use this new feature though: https://mariadb.org/grant-to-public-in-mariadb/) So I think we can move on.

plugins/modules/mysql_db.py
laurent-indermuehle: @noonedeadpunk  I'm not well rested... you're right. Fixing now! Good job catching that!

laurent-indermuehle: I dislike the idea of setting server_implementation as "mysql" for older version of MariaDB. This could add confusion while debugging.

I pushed a version that bother about version but I think it's not that complex. What do you think @noonedeadpunk?

noonedeadpunk (approved):

plugins/modules/mysql_db.py
noonedeadpunk: oh, yes, that is much better then what I suggested :)

noonedeadpunk (approved):

Andersson007 (approved):
@laurent-indermuehle LGTM, feel free to merge when you think it's ready

laurent-indermuehle:
Thanks for your review @Andersson007! I'm merging now

Andersson007:
@laurent-indermuehle @noonedeadpunk thanks for the contribution!

Andersson007:
@laurent-indermuehle should we also update our README here https://github.com/ansible-collections/community.mysql#databases?

laurent-indermuehle:
What do you mean @Andersson007? I think I did by adding this:
```md
mariadb:10.5.25 (collection version >= 3.10.0, <3.13.0)
[..-]
mariadb:11.4.5 (collection version >= 3.13.0)
```
Did I missed something else?

Andersson007:
> What do you mean @Andersson007? I think I did by adding this:
> 
> ```
> mariadb:10.5.25 (collection version >= 3.10.0, <3.13.0)
> [..-]
> mariadb:11.4.5 (collection version >= 3.13.0)
> ```
> 
> Did I missed something else?

Ah, cool, missed this somehow in the PR.. sorry for the noise and thanks!

PR #712, opened 2025-04-24
fix mysql_info crash if PUBLIC role is used on MariaDB 10.11+

laurent-indermuehle:
##### SUMMARY

MariaDB 10.11 introduced a PUBLIC role: https://mariadb.com/kb/en/grant/#to-public

Because roles in MariaDB have an empty "host" column in the users table, the mysql_info module crash with the error `ERROR 1141, There is no such grant defined for user 'PUBLIC' on host '%'`.  This PR fix that by detecting if the host is empty, then it uses the  `privileges_get` method with the `maria_role` argument sets to True.

This fix addresses the crash but doesn't add comprehensive support for roles. As explained in the changelog fragment, retrieving both direct privileges and inherited role privileges for each user would require significant module refactoring. It's worth noting that this limitation mirrors MariaDB's own CLI behavior, which also doesn't display comprehensive privilege information for users with roles. Users should be aware of these limitations when using MariaDB roles with Ansible.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

mysql_info


laurent-indermuehle:
The failed tests are from devel (Ansible 2.19) that changed its handling of variable types and is not related to this PR.
Ready for review!

Andersson007 (approved):

laurent-indermuehle:
I rebased onto upstream/main to retrieve the CI changes from #713.
I hope the test will all pass now.

laurent-indermuehle:
Green! I merge. Thanks for the review @Andersson007 !

PR #715, opened 2025-05-05
Add query to find command between python and python3

laurent-indermuehle:
##### SUMMARY

I noticed that the command `python -V` fails in integrations tests. Python is installed by https://github.com/ansible-community/ansible-test-gh-action so the test suite from `community.mysql` doesn't control what version is installed and if an alias from `python` to `python3` is added.

So this PR fix the check that verify what version of Python is present inside the test container/VM.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

CI

### DIFF
--- a/tests/integration/targets/setup_controller/tasks/verify.yml
+++ b/tests/integration/targets/setup_controller/tasks/verify.yml
@@ -42,8 +42,8 @@
         - connector_name == 'mysqlclient'
 
     - name: Get the python version in use
-      ansible.builtin.command:
-        cmd: python -V
+      ansible.builtin.shell:
+        cmd: echo $( $(command -v python || command -v python3) -V )
       changed_when: false
       failed_when: false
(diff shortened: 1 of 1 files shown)


laurent-indermuehle:
Hi @Andersson007. Please, I'm ready for a review. This is not much, but it's one error gone in all runs!

In test's artifacts, if you search for `command -v` you will see:

```
TASK [setup_controller : Display the python version in use] ********************
ok: [testhost] => {
    "msg": "Python in use inside the test container: Python 3.10.12\n"
}
```
Which by chance correspond to what is said to be used in the readme. But this is a happy accident as we rely on the default from the github action ansible-test-gh-action for this. I'm not sure if it easy or a good idea to pin a version in the integrations tests.

Andersson007 (approved):

Andersson007:

PR #740, opened 2025-09-16
CI: remove testing against devel from matrix for stable-3 branch

Andersson007:
##### SUMMARY
CI: remove testing against devel from matrix for stable-3 branch

It's not a requirement for included collection to test against devel, so lets remove it from stable-3 to avoid crashes related to Python 3 as Python 2 is still supported in stable-3


laurent-indermuehle (approved):
LGTM

PR #770, opened 2026-02-12
ci: add ingore.txt entry to avoid failures with core 2.21

Andersson007:
##### SUMMARY
ci: add ingore.txt entry to avoid failures with core 2.21

i investigated a lot: we can't use run_command instead of Popen because it doesn't provide a pipe that will transmit data from a dump to stdin of mysql


Andersson007:
ready for review

PR #779, opened 2026-03-09
Add parsec to ed25519 plugin check

domulink:
##### SUMMARY
When creating a new user, and specifying parsec as plugin it will generate the wrong SQL query.
It should be added to the same plugin check as ed25519 so that it generates a query using` USING PASSWORDS(%s)`
instead of `BY %s`

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`community.mysql.mysql_user`

##### ADDITIONAL INFORMATION
MariaDB has intended parsec to be the default in future releases. It is avaiable since MariaDB 11.6 the docs for reference:
[ https://mariadb.com/docs/server/reference/plugins/authentication-plugins/authentication-plugin-parsec](https://mariadb.com/docs/server/reference/plugins/authentication-plugins/authentication-plugin-parsec)


Andersson007 (commented):
@domulink thanks for the fix! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)?

PR #774, opened 2026-02-18
Fix deprecated imports

vphammer:
##### SUMMARY

Updates imports to resolve #771 and address deprecation of ansible.module_utils.six.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Module imports
##### ADDITIONAL INFORMATION

This change will make this collection incompatible with managed hosts running Python 2.7.

PR #735, opened 2025-09-15
Announce dropping Python 2 support

Andersson007:
##### SUMMARY

Announce dropping Python 2 support

PR #781, opened 2026-03-16
Release 4.2.0 commit

Andersson007:
##### SUMMARY
Release 4.2.0 commit

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,30 @@ Community MySQL and MariaDB Collection Release Notes
 
 This changelog describes changes after version 2.0.0.
 
+v4.2.0
+======
+
+Release Summary
+---------------
+
+This is a minor release of the ``community.mysql`` collection.
+This changelog contains all breaking changes to the modules in this collection
+
+Minor Changes
+-------------
+
+- Update imports from ansible.module_utils._text to use ansible.module_utils.common.text.converters
+
+Breaking Changes / Porting Guide
+--------------------------------
+
+- Update imports from ansible.module_utils.six to use their python3 equivalent. This change will make this collection incompatible for managed hosts on python2.7.
+
+Bugfixes
+--------
+
+- mysql_user - When creating a new user, and specifying parsec as plugin it wil generate the wrong SQL query. It should be added to the same plugin check as ed25519 so that it generates a query using USING PASSWORDS(%s) instead of BY %s (https://github.com/ansible-collections/community.mysql/pull/779).
+
 v4.1.0
 ======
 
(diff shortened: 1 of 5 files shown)


mariolenz:
@Andersson007 A breaking change like #774 really [shouldn't](https://docs.ansible.com/projects/ansible/latest/community/collection_contributors/collection_requirements.html#versioning-and-deprecation) have been done in a [minor release](https://github.com/ansible-collections/community.mysql/releases/tag/4.2.0) :smirk:

cc @vphammer

Andersson007:
@mariolenz hi, please see https://github.com/ansible-collections/community.mysql/pull/774#issuecomment-3925897716

As the collection supports Python 3 since 4.0.0, I believe the PR is valid. Users [were notified in advance](https://github.com/ansible-collections/community.mysql/pull/735/changes).
If your concern is only about the `breaking_changes` key we used in changelog, I'm not sure what we should use instead, maybe `minor_changes`? The intention was to bring users' attention - maybe it's an extra thing and kinda implied since 4.0.0 was released. WDYT?

mariolenz:
Ah, yes... now I see it! The collection already dropped support for Python 2 in [4.0.0](https://github.com/ansible-collections/community.mysql/releases/tag/4.0.0). So this _isn't_ really a breaking change in a minor release. I've been confused to see breaking changes in a minor release, especially one that's already included in the Ansible Community Package.

> If your concern is only about the `breaking_changes` key we used in changelog, I'm not sure what we should use instead, maybe `minor_changes`? The intention was to bring users' attention - maybe it's an extra thing and kinda implied since 4.0.0 was released. WDYT?

I'm not sure. But between `minor_changes` and `breaking_changes` there are also [major_changes](https://docs.ansible.com/projects/ansible/devel/community/development_process.html#creating-a-changelog-fragment). Major changes should, I hope, also bring users' attention. Although it's somewhat unfortunate that major _changes_ don't require a major _release_, I always find this confusing. Maybe `important_changes` or something might have been better. But this looks like an ansible-core decision to me.

Well, sorry for the confusion. I'm with you, it's not a breaking change. Shouldn't show up as one in the changelog, though, in that case. But I can live with this.

Andersson007:
@mariolenz what do you think about starting a community-wg forum topic on this? Feel free to create or I could if you have no time. Please let me know

mariolenz:
> @mariolenz what do you think about starting a community-wg forum topic on this? Feel free to create or I could if you have no time. Please let me know

[Clarification on breaking changes in changelogs](https://forum.ansible.com/t/45711)

Andersson007:
@mariolenz thanks a lot for creating the topic!

PR #782, opened 2026-03-17
bugfix: handle non-JSON-serializable types (Decimal, datetime) in mysql_query results

### LINKED ISSUE #783: mysql_query fails with "Value of unknown type: <class 'decimal.Decimal'>" for DECIMAL/DATETIME columns
## Bug description

`mysql_query` fails with the following error when a query returns columns of types that are not natively JSON-serializable, such as `DECIMAL`/`NUMERIC` (returned as `decimal.Decimal`) or `DATE`/`DATETIME`/`TIMESTAMP` (returned as `datetime`):

```
"msg": "Task failed: Module failed: Value of unknown type: <class 'decimal.Decimal'>, -18.51"
```

## Steps to reproduce

```sql
CREATE TABLE test_decimal (id INT PRIMARY KEY, price DECIMAL(10,2));
INSERT INTO test_decimal VALUES (1, 18.51);
```

```yaml
- name: Query a DECIMAL column
  community.mysql.mysql_query:
    login_db: mydb
    query: SELECT id, price FROM test_decimal
```

**Result:** task fails with `Value of unknown type: <class 'decimal.Decimal'>, 18.51`

**Expected:** task succeeds and returns the value (as a string or number)

## Root cause

`cursor.fetchall()` returns raw Python objects. `decimal.Decimal`, `datetime.datetime`, `datetime.date`, `datetime.timedelta`, and `bytes` are not natively JSON-serializable. Ansible's `exit_json` calls `json.dumps` internally, which raises `TypeError: Object of type Decimal is not JSON serializable`, surfaced as `Value of unknown type`.

The affected line in `plugins/modules/mysql_query.py`:

```python
query_result.append([dict(row) for row in cursor.fetchall()])
```

## Environment

- community.mysql version: 3.10.3 (also present in latest `main` as of March 2026)
- MySQL driver: pymysql
- Affected column types: `DECIMAL`, `NUMERIC`, `DATE`, `DATETIME`, `TIMESTAMP`, `TIME`, `BLOB`, `BINARY`

## Suggested fix

Round-trip through `json.dumps(default=str)` to coerce non-serializable types to their string representations before passing to `exit_json`:

```python
rows = [dict(row) for row in cursor.fetchall()]
query_result.append(json.loads(json.dumps(rows, default=str)))
```

A fix and integration test are proposed in PR #782.

ekbritecore:
Fixes #783

## Summary

`mysql_query` fails with the following error when a query returns columns with types that are not natively JSON-serializable (e.g. `decimal.Decimal`, `datetime`, `bytes`):

```
"msg": "Task failed: Module failed: Value of unknown type: <class 'decimal.Decimal'>, -18.51"
```

This happens because `cursor.fetchall()` returns raw Python objects that Ansible's `exit_json` then tries to serialize to JSON directly — and fails on types MySQL drivers return as `decimal.Decimal` (e.g. `DECIMAL`, `NUMERIC` columns) or `datetime` (e.g. `DATE`, `DATETIME`, `TIMESTAMP` columns).

## Fix

Round-trip the fetched rows through `json.dumps(default=str)` + `json.loads` before appending to `query_result`. This coerces any non-serializable value to its string representation, which is the correct and safe behavior.

```python
# Before
query_result.append([dict(row) for row in cursor.fetchall()])

# After
rows = [dict(row) for row in cursor.fetchall()]
query_result.append(json.loads(json.dumps(rows, default=str)))
```

This pattern has been the standard fix for this class of problem (see prior art in a similar mssql module: https://github.com/melmorabity/ansible-mssql-query/issues/2).

## Affected types

The following MySQL column types are known to trigger this:

| Column type | Python type returned |
|-------------|---------------------|
| `DECIMAL`, `NUMERIC` | `decimal.Decimal` |
| `DATE`, `DATETIME`, `TIMESTAMP` | `datetime.date` / `datetime.datetime` |
| `TIME` | `datetime.timedelta` |
| `BLOB`, `BINARY` | `bytes` |

## Changes

- `plugins/modules/mysql_query.py` — add `import json`, replace raw `fetchall` append with JSON round-trip
- `tests/integration/targets/test_mysql_query/tasks/issue-783.yml` — integration test: creates a table with `DECIMAL` and `DATETIME` columns, inserts a row, queries it, asserts success and that values are returned as strings
- `tests/integration/targets/test_mysql_query/tasks/main.yml` — include new test
- `changelogs/fragments/mysql_query_decimal_serialization.yaml` — changelog entry


Andersson007 (approved):

Andersson007:
@ekbritecore thanks for the contribution!
I've merged it because the CI failures were unrelated.
I created an issue https://github.com/ansible-collections/community.mysql/issues/784, help there would be much appreciated

PR #819, opened 2026-05-29
fix(mysql_user): add MySQL 9.7.0 compatibility for password management

Andersson007:
## Summary

Fix `mysql_user` failures on MySQL 9.7.0+ by version-gating password management to use `IDENTIFIED BY` instead of client-side SHA1 hashing with `mysql_native_password`.

## Problem

Two breaking changes in MySQL 9.x prevent `mysql_user` from working:

1. **`mysql_native_password` removed** (MySQL 9.0+). The module uses `CREATE USER ... IDENTIFIED WITH mysql_native_password AS ...` and `ALTER USER ... IDENTIFIED WITH mysql_native_password AS ...` for all password operations. These fail on 9.0+.

2. **`SHA1()` SQL function removed from server core** (MySQL 9.6+). The module calls `SELECT CONCAT('*', UCASE(SHA1(UNHEX(SHA1(...)))))` to hash passwords client-side and detect changes. This fails with `FUNCTION SHA1 does not exist` on 9.6+.

MySQL 9.7.0 is the first LTS release in the 9.x series, where both problems converge.

## Solution

All changes are version-gated at 9.7.0. Older MySQL and MariaDB behavior is unchanged.

- **`user_add()`**: Use `CREATE USER %s@%s IDENTIFIED BY %s` on 9.7.0+, letting the server hash with `caching_sha2_password`.
- **`user_mod()`**: Use `ALTER USER %s@%s IDENTIFIED BY %s` on 9.7.0+, skipping hash comparison (impossible with random-salt hashing — see below).
- **`encrypted=True`**: `fail_json` on 9.7.0+ with an actionable message, since `mysql_native_password` hashes are unusable.
- **Tests**: Version-guard assertions using two thresholds: 9.7 for password idempotency (random-salt issue), 9.0 for `mysql_native_password` plugin removal. Guards added across `test_mysql_user`, `test_mysql_info`, and `test_mysql_replication`. Unit tests added for `server_supports_mysql_native_password()`.

### Why always-changed on 9.7.0+

`caching_sha2_password` uses a [20-byte random salt with 5000 rounds of SHA-256](https://dev.mysql.com/blog-archive/mysql-8-0-4-new-default-authentication-plugin-caching_sha2_password/). The same plaintext password produces a different hash every time, making client-side hash comparison impossible. The module reports `changed=true` whenever a password is provided with `update_password: always` (the default). Users who need idempotency can use `update_password: on_create`.

## Documentation references

- [mysql_native_password removed in MySQL 9.0](https://dev.mysql.com/doc/relnotes/mysql/9.0/en/news-9-0-0.html)
- [SHA1()/MD5() relocated to optional component in MySQL 9.6](https://dev.mysql.com/doc/relnotes/mysql/9.7/en/news-9-6-0.html)
- [caching_sha2_password reference](https://dev.mysql.com/doc/refman/8.4/en/caching-sha2-pluggable-authentication.html)
- [MySQL 9.7.0 LTS release notes](https://dev.mysql.com/doc/relnotes/mysql/9.7/en/news-9-7-0.html)

## Alternatives considered

| Approach | Why rejected |
|---|---|
| Install `component_classic_hashing` to restore SHA1() | Module shouldn't modify server config; `mysql_native_password` is still gone |
| Compute `caching_sha2_password` hashes client-side | Random salt makes comparison impossible; no public API for the hash format |
| Authenticate as the target user to detect password changes | Security risks (lockout policies, connection restrictions), performance overhead |

`IDENTIFIED BY` is [MySQL's own recommended approach](https://dev.mysql.com/doc/refman/9.7/en/alter-user.html) for setting passwords on 9.x — the server handles hashing internally.

## Known limitations

- `mysql_user` always reports `changed=true` when a password is provided on MySQL 9.7.0+ with `update_password: always`.
- `encrypted` option is not supported on MySQL 9.7.0+.
- The module code version-gates at 9.7.0 (first LTS). MySQL 9.0–9.6 (Innovation releases) are not covered by the module gate, though test guards do cover 9.0+ for plugin-specific tests.
- Replication integration tests (`test_mysql_replication`) skip IO-thread and data-replication assertions on MySQL 9.0+ because `caching_sha2_password` requires SSL or `GET_SOURCE_PUBLIC_KEY` for the replication handshake, and the test containers do not configure either. TODO: configure SSL for test containers to restore full replication coverage on 9.0+.

Assisted-by: Claude Opus

### DIFF
--- a/plugins/module_utils/user.py
+++ b/plugins/module_utils/user.py
@@ -217,10 +217,17 @@ def user_add(cursor, user, host, host_all, password, encrypted,
     if password and encrypted:
         if impl.supports_identified_by_password(cursor):
             query_with_args = "CREATE USER %s@%s IDENTIFIED BY PASSWORD %s", (user, host, password)
-        else:
+        elif impl.server_supports_mysql_native_password(cursor):
             query_with_args = "CREATE USER %s@%s IDENTIFIED WITH mysql_native_password AS %s", (user, host, password)
+        else:
+            # MySQL 9.7.0+: encrypted hashes are mysql_native_password format, unusable.
+            module.fail_json(msg="The 'encrypted' option is not supported on MySQL 9.7.0+ "
+                                 "because the mysql_native_password plugin has been removed. "
+                                 "Use a plaintext password instead.")
     elif password and not encrypted:
-        if old_user_mgmt:
+        # MySQL 9.7.0+: SHA1() and mysql_native_password were removed;
+        # let the server hash with caching_sha2_password via IDENTIFIED BY.
+        if old_user_mgmt or not impl.server_supports_mysql_native_password(cursor):
             query_with_args = "CREATE USER %s@%s IDENTIFIED BY %s", (user, host, password)
         else:
             cursor.execute("SELECT CONCAT('*', UCASE(SHA1(UNHEX(SHA1(%s)))))", (password,))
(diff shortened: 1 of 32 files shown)


Andersson007:
cc @laurent-indermuehle @betanummeric @eRadical @ronger4 I'd appreciate thorough looks at this PR

this PR also contains a lot of changes to tests unrelated to mysql_user, but to their incompatibility with 9.7.0 in general

plugins/module_utils/user.py
```diff
@@ -304,69 +311,82 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
...
-                    else:
-                        cursor.execute("SELECT CONCAT('*', UCASE(SHA1(UNHEX(SHA1(%s)))))", (password,))
-                    encrypted_password = cursor.fetchone()[0]
-
-                if current_pass_hash != encrypted_password:
+                if not impl.server_supports_mysql_native_password(cursor):
+                    # MySQL 9.7.0+: mysql_native_password and SHA1() are removed.
+                    # We cannot compare password hashes (caching_sha2_password
+                    # uses a random salt), so always update the password.
+                    if encrypted:
```
laurent-indermuehle, line 318: This test will return an MySQL 9.7 error on MariaDB when it will stop supporting mysql_native_password too. Maybe a more generic error message is best?

laurent-indermuehle (approved):
SGTM, but wow, this is big. Just for this, I'll vote for a split in https://forum.ansible.com/t/should-ansible-mysql-be-split-into-separate-mysql-and-mariadb-collections/45866

plugins/module_utils/user.py
```diff
@@ -304,69 +311,82 @@ def user_mod(cursor, user, host, host_all, password, encrypted,
...
+                        module.fail_json(msg="The 'encrypted' option is not supported on MySQL 9.7.0+ "
+                                             "because the mysql_native_password plugin has been removed. "
+                                             "Use a plaintext password instead.")
```
```suggestion
                        module.fail_json(msg="The 'encrypted' option is not supported by your database server "
                                             "version because the mysql_native_password plugin has been removed. "
                                             "Use a plaintext password instead.")
```

plugins/module_utils/user.py
Andersson007: @laurent-indermuehle makes sense, how about https://github.com/ansible-collections/ansible.mysql/pull/819#discussion_r3339245776 ?

Andersson007:

PR #820, opened 2026-06-04
docs(README.md): update versions of software we test against

Andersson007:
##### SUMMARY
Make README correspond to https://github.com/ansible-collections/ansible.mysql/blob/main/.github/workflows/ansible-test-plugins.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ Here is the table for the support timeline:
 
 ### ansible-core
 
+- stable-2.16 (RHEL8 context only)
 - stable-2.17
 - stable-2.18
 - stable-2.19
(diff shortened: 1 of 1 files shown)


laurent-indermuehle (approved):
Good catch!

maybe we can use only minor version: 8.0 instead of 8.0.38 to avoid having to update this everytime we bump ours tests? We did that for MariaDB already.

Andersson007:
> Good catch!
> 
> maybe we can use only minor version: 8.0 instead of 8.0.38 to avoid having to update this everytime we bump ours tests? We did that for MariaDB already.

@laurent-indermuehle i was thinking about the same but wanted to raise this in a dedicated PR (we should check out code, do a proper anouncment, etc.)
Thanks for reviewing!

Issue #802, opened 2026-05-11
MySQL High Availability and Clustering Modules (InnoDB Cluster, Group Replication, MySQL Router)


ronger4:
### Summary

Hey everyone,
I'd like to propose adding MySQL High Availability support to the collection
and would love your feedback before I start building anything.
### The problem
Right now the collection handles basic primary-replica replication via
mysql_replication, but it doesn't cover MySQL's modern HA stack - InnoDB
Cluster, Group Replication, and MySQL Router. If you need to automate any of
that today, you're stuck shelling out to mysqlsh or writing raw SQL in
command tasks. That's fragile, hard to make idempotent, and no fun to
maintain.
### What to consider
I'm considering three new module groups:
**1. mysql_innodb_cluster / mysql_innodb_cluster_info**
Manage the full InnoDB Cluster lifecycle, create a cluster, add/remove nodes,
rejoin after failure, dissolve, check status. The info module would return
topology, member health, and replication lag so you can use it in playbooks.
**2. mysql_group_replication**
For deployments that use Group Replication directly without the AdminAPI
wrapper. Start/stop GR on individual nodes, bootstrap a group, switch between
single-primary and multi-primary mode, elect a new primary. This would live
alongside mysql_replication, not replace it.
**3. mysql_router / mysql_router_info**
Bootstrap MySQL Router against a cluster, configure routing (read/write
splitting, connection pooling, TLS), and query Router status.
### Things worth discussing

**Should Group Replication be its own module?**
It overlaps with mysql_replication conceptually. 
**Module granularity**
Is one mysql_innodb_cluster module with an action parameter the right
design, or would you rather see finer-grained modules
(mysql_innodb_cluster_member, etc.)?
**Minimum MySQL version**
InnoDB Cluster requires MySQL 8.0+. Should we just target 8.0+ ?
**MariaDB Galera**
I'm intentionally leaving Galera out of this proposal, it's a different
clustering solution with different tooling. 
### What's out of scope
- NDB Cluster (completely different architecture)
- ProxySQL (third-party, separate discussion)
- Installing MySQL Server, MySQL Shell, or MySQL Router packages

We're at the "thinking about it" stage, nothing is set in stone. If you have
use cases, concerns, alternative ideas -  I'd appreciate it.

### Issue Type

Feature Idea

### Component Name

mysql_innodb_cluster, mysql_innodb_cluster_info, mysql_router, mysql_router_info

### Additional Information

```yaml (paste below)

```

[...]

PR #821, opened 2026-06-04
feat: add mysql_innodb_cluster and mysql_innodb_cluster_info modules

ronger4:
#### SUMMARY
- New `mysql_innodb_cluster` module for managing InnoDB Cluster lifecycle (create, dissolve, add/remove instances, set_primary, switch topology, rescan, set_option)
- New `mysql_innodb_cluster_info` module for querying cluster status, members, options, and routers
- Shared `module_utils/mysqlsh.py` wrapper for MySQL Shell CLI integration
- New `doc_fragments/mysql_innodb_cluster.py` for mysqlsh connection parameters
- Dedicated Makefile target (`test-integration-innodb-cluster`) and CI job for InnoDB Cluster tests
- Unit tests for all new components
- Integration tests covering cluster lifecycle with idempotency and check_mode

#### ISSUE TYPE
- New Module Pull Request

#### COMPONENT NAME
- plugins/modules/mysql_innodb_cluster.py
- plugins/modules/mysql_innodb_cluster_info.py
- plugins/module_utils/mysqlsh.py
- plugins/doc_fragments/mysql_innodb_cluster.py
- Makefile
- .github/workflows/ansible-test-plugins.yml
- TESTING.md

#### ADDITIONAL INFORMATION
- InnoDB Cluster tests are separated from regular tests due to GTID/GR requirements conflicting with mysql_db dump/import
- Makefile refactored with shared macros to reduce duplication
- PR was created by the help of Cursor + Opus 4.6 with Human review, manual tests and feedback.
-  Implements one part of #802
- Worth refactoring the makefile and CI parts that are duplicated both in the regular legacy tests and new tests, didn't want to add it now as it will make the PR bigger than its already is


laurent-indermuehle:
Hi @ronger4, it's not clear if this is a draft or ready for review. GH help you convert to a draft in the upper right corner.

Let us know if you need help with the failing CI tests.

ronger4:
> Hi @ronger4, it's not clear if this is a draft or ready for review. GH help you convert to a draft in the upper right corner.
> 
> Let us know if you need help with the failing CI tests.

This is still a draft, my bad not marking it.
It is in an advanced stage once done I will ping for review.

Currently tests fail due to new CI changes to support innodb part, I will explain the motivation once things get green

ronger4:
@laurent-indermuehle The PR is ready for a review.
Few notes to take into consideration:

-  Innodb cluster feature couldn't be implemented by using the pymsql that implements the other modules, the way to implement it is by using mysqlsh. Util wrapper for that functionality was introduced here.
- Setting up integration tests for innodb cluster require some pre-config like GTID and group replication settings. This conflicts and breaks some of the tests of the original modules like dump/import. Because of that I had to split those tests to 2 different makefile functions and trigger them separately.
- If there are suggestions for improvements or ideas for better directions I'd be happy to hear and take that forward.

laurent-indermuehle (changes requested):
It's unfortunate that a new connector, new test infra, new test workflow and new Makefile had to be added. It add complexity. Even the doc_framgments is mostly a duplicate.

But all in all, it seams to me there are necessary changes and that you have done a great job to make this new module works and been tested!

The issue I see and I would like to see resolved before approving: In the checks, every integrations tests ran. Even for MariaDB. Do you think you could modify the `on` close of the pipelines so that only the appropriate tests run?

ronger4:
@laurent-indermuehle I felt the same while working on this PR. I tried avoiding the creation of extra tests in the makefile and this split but no matter what I tried I couldn't find a more clean solution unfortunately (I'm open for suggestions if one comes up to mind). I will work on an extra PR after that one that would address duplicates and will try put those duplicates maybe to a common script that can be shared.

In addition maybe in the same PR I will add the support for mysql 9.7 that was added recently, it feels too much to add it here as-well. PR is complex enough as is.

About the last common, from my understanding every PR runs all the tests right?
In this case innodb specific tests run only on relevant mysql versions, its not running on mariadb.
Or the expectation in general is to have a mechanism that runs the innodb tests only on PRs that have changes related to mysql innodb part of the code for example? It feels like a new feature that is not present. Or did I misunderstood?

laurent-indermuehle:
@ronger4 thanks for your honesty and completeness. I also think the split was the best way.

About my last comment, I was referring to this part:
```yaml
name: Plugins CI
on:  # yamllint disable-line rule:truthy
  push:
    paths:
      - 'plugins/**'
      - 'tests/**'
      - '.github/workflows/ansible-test-plugins.yml'
  pull_request:
    paths:
      - 'plugins/**'
      - 'tests/**'
      - '.github/workflows/ansible-test-plugins.yml'
```
I think we need to to yet another split : By adding a new workflow. So in addition to this workflow:
- .github/workflows/ansible-test-plugins.yml
We could add a separate one:
- .github/workflows/ansible-test-plugins-innodb-cluster.yml

Tests are expensives! They use a lot of VMs, runs for a long time and produce a tone of output. So the goal is to run them only if relevant files changes. I don't believe every PR should run all tests. Please @Andersson007 correct me if I'm wrong.

Unless, you think it will be necessary to test the others modules against this new topology? For instance:
- Is it possible to use mysql_query or mysql_variables on a innodb-cluster ?
- Is it possible to take a backup with mysql_db ?
If so, what about others topologies? Galera? Group Replication?
But I don't think it is necessary. A query or a logical backup should run the same way on a standalone or a cluster.

ronger4:
After discussing that further with the community in the community matrix chat, it will be dropped for now as this adds complexity and divergence from the main modules

Andersson007:
@ronger4 I'll close the PR then to keep the tracker clean. Anyway, thank you folks much for the effort and the discussion!

Issue #803, opened 2026-05-11
Adding MySQL Backup, Recovery, and Binary Log Management modules


ronger4:
### Summary

Hey everyone,
I'd like to propose adding proper backup, recovery, and binary log management
support to the collection and would love your feedback.

### The problem

The collection has `mysql_db` with `state: dump` and `state: import`, which
covers basic mysqldump/import. But that's about where it stops. If you need
encrypted or compressed backups, binary log purging, point-in-time recovery,
or fast node provisioning via mysql's Clone plugin — you are writing shell
tasks and managing all the edge cases yourself.

### What I'm considering

**1. mysql_backup**
Go beyond what mysql_db does today. Support parallel dumps (via mysqlpump
or mysql Shell's dump utilities), encryption, compression (zstd, gzip),
selective backup (specific databases/tables including routines and triggers),
and backup validation (restore-and-verify). Also retention cleanup — delete
backups older than N days.

This wouldn't replace mysql_db for simple dump/import . It would be a
dedicated module for more serious backup workflows.

**2. `mysql_binlog` / `mysql_binlog_info`**
Manage the binary log lifecycle: configure format (ROW/MIXED/STATEMENT),
set expiration, purge old logs by date or filename, and enable binlog
encryption. The info module would return the current log file, position,
list of all binary logs, and disk usage are useful for planning PITR and
monitoring disk consumption.

**3. `mysql_clone`**
Wrap mysql 8.0.17+ Clone plugin. Initiate a remote clone from a donor
instance, monitor progress via `performance_schema`, configure the valid
donor list and SSL options. This is the fastest way to provision a new
replica or rebuild a failed node. A single operation instead of a full
logical dump/restore.

### Things worth discussing

**`mysql_backup` vs `mysql_db`**
`mysql_db` already does dump/import. Should `mysql_backup` be a completely
separate module (cleaner separation, enterprise-focused features), or should
we extend `mysql_db` with new options? Extending avoids a new module but
risks overloading an already busy one.

**Binary log config: new module or `mysql_variables`?**
You can already set `binlog_format` and `binlog_expire_logs_seconds` via
`mysql_variables`. Does a dedicated `mysql_binlog` module add enough value
(purging, info gathering, grouping related settings), or is it unnecessary?

**Clone plugin scope**
Should `mysql_clone` also handle installing the clone plugin itself
(`INSTALL PLUGIN clone ...`), or assume it's already installed? The plugin
[...]

PR #824, opened 2026-06-11
Add mysql_binlog_info module

ronger4:
#### SUMMARY
- add a new read-only `mysql_binlog_info` module for inspecting binary log status, files, totals, and selected binlog settings
- normalize current binlog status across MySQL and MariaDB, including MariaDB GTID fallback behavior

#### ISSUE TYPE
- New Module Pull Request

#### COMPONENT NAME
- `TESTING.md`
- `meta/runtime.yml`
- `plugins/modules/mysql_binlog_info.py`
- `tests/integration/targets/test_mysql_binlog_info/defaults/main.yml`
- `tests/integration/targets/test_mysql_binlog_info/meta/main.yml`
- `tests/integration/targets/test_mysql_binlog_info/tasks/main.yml`
- `tests/unit/plugins/modules/test_mysql_binlog_info.py`

#### ADDITIONAL INFORMATION
- the module is intentionally read-only; mutable binlog settings remain with `mysql_variables`
- Code was created with the help for GPT5.4 using Cursor with Human in the loop and manual tests.
- Impleemnts #803 partially

### DIFF
--- a/plugins/modules/mysql_binlog_info.py
+++ b/plugins/modules/mysql_binlog_info.py
@@ -1,0 +1,327 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: mysql_binlog_info
+
+short_description: Gather MySQL or MariaDB binary log information
+
+description:
+  - Gather binary log information from MySQL or MariaDB servers.
+  - Returns the current binary log status, available binary log files, aggregate log totals,
+    and selected binary-log-related settings.
+  - This module is read-only. Use M(ansible.mysql.mysql_variables) for mutable binlog settings.
+
+author:
+  - Ron Gershburg (@rgershbu)
+
+version_added: '5.1.0'
+
+options:
+  filter:
+    description:
+      - Limit the collected information by comma separated string or YAML list.
+      - Allowable values are C(current), C(logs), C(totals), and C(settings).
+      - By default, collects all subsets.
+      - You can use C(!) before a value, for example C(!settings), to exclude it from the information.
+      - If you pass including and excluding values together, the excluding values are ignored.
+    type: list
+    elements: str
+    version_added: '5.1.0'
+
+notes:
+  - Compatible with MariaDB or MySQL.
+  - The module does not modify server state.
+  - Some mutable binlog settings are dynamic and can be managed with M(ansible.mysql.mysql_variables).
+  - Startup-only or restart-managed settings remain outside this module's scope.
+
+attributes:
+  check_mode:
+    support: full
+  idempotent:
+    support: full
+
+seealso:
+  - module: ansible.mysql.mysql_variables
+  - module: ansible.mysql.mysql_replication
+
+extends_documentation_fragment:
+  - ansible.mysql.mysql
+'''
+
+EXAMPLES = r'''
+- name: Collect all binlog information
+  ansible.mysql.mysql_binlog_info:
+    login_user: root
+    login_unix_socket: /run/mysqld/mysqld.sock
+
+- name: Collect only current binlog status and log list
+  ansible.mysql.mysql_binlog_info:
+    login_user: root
+    login_password: rootpass
+    filter:
+      - current
+      - logs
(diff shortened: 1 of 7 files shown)


ronger4:
@eRadical @Andersson007 @laurent-indermuehle can you review please?

I think the CI failure is a fluke, worth re-running this specific job. It passes locally for me (the exact parameters)

plugins/modules/mysql_binlog_info.py
```diff
@@ -0,0 +1,327 @@
...
+  filter:
+    description:
+      - Limit the collected information by comma separated string or YAML list.
+      - Allowable values are C(current), C(logs), C(totals), and C(settings).
+      - By default, collects all subsets.
+      - You can use C(!) before a value, for example C(!settings), to exclude it from the information.
+      - If you pass including and excluding values together, the excluding values are ignored.
+    type: list
+    elements: str
+    version_added: '5.1.0'
```
Andersson007:
```suggestion
```
When we add a new module, we only need to have `version_added` at the root level of the DOCUMENTATION block

plugins/modules/mysql_binlog_info.py
```diff
@@ -0,0 +1,327 @@
...
+author:
+  - Ron Gershburg (@rgershbu)
+
+version_added: '5.1.0'
+
+options:
+  filter:
+    description:
+      - Limit the collected information by comma separated string or YAML list.
+      - Allowable values are C(current), C(logs), C(totals), and C(settings).
```
Andersson007:
```suggestion
      - Allowable values are V(current), V(logs), V(totals), and V(settings).
```

plugins/modules/mysql_binlog_info.py
```diff
@@ -0,0 +1,327 @@
...
+      - By default, collects all subsets.
+      - You can use C(!) before a value, for example C(!settings), to exclude it from the information.
```
Andersson007:
```suggestion
      - You can use C(!) before a value, for example V(!settings), to exclude it from the information.
```

Andersson007 (commented):
@ronger4 thanks for submitting a new module!

Andersson007 (approved):
@ronger4 LGTM, thanks!

Next time, I would appreciate if my suggestions are incorporated by using "Commit" button in corresponding comments so that they are counted as my contributions too. Not a big deal though:)

ronger4:
> @ronger4 LGTM, thanks!
> 
> Next time, I would appreciate if my suggestions are incorporated by using "Commit" button in corresponding comments so that they are counted as my contributions too. Not a big deal though:)

Sorry for not doing that, I don't tend to work like that. Will do next time :)
Thank you for the review!

can you merge please? I don't have the permissions

laurent-indermuehle (changes requested):
Thanks @ronger4, this is well writen and it supports both MySQL and MariaDB, couldn't be more happy!

Just one thing: What happen when the server is running with log_bin: false?
I didn't see a test (in the module) or an integration test for this case I think.

ronger4:
> Thanks @ronger4, this is well writen and it supports both MySQL and MariaDB, couldn't be more happy!
> 
> Just one thing: What happen when the server is running with log_bin: false? I didn't see a test (in the module) or an integration test for this case I think.

This is a valid point. Currently I think it will fall.
There are 2 options that we can handle that.
1) Return an empty list - it makes sense because there are no bin logs so the returned state is correct.
2) Return error message saying it log_bin is off.

@laurent-indermuehle  @Andersson007  WDYT?

laurent-indermuehle:
explicit is better, always :) I'm in favor of a warning or failure.
I think it will depend on the goal of the module. If it's only a warning, subsequent tasks will fails with a misleading error message. While a failure will stop the playbook at the right place.

Maybe #803 can answer that?

ronger4:
> explicit is better, always :) I'm in favor of a warning or failure. I think it will depend on the goal of the module. If it's only a warning, subsequent tasks will fails with a misleading error message. While a failure will stop the playbook at the right place.
> 
> Maybe #803 can answer that?

@laurent-indermuehle Thanks, I added explicit failure. I wrote unit tests but I didn't write integration test for that because it would require changing binlog to off on container level and it might break the other integration suite tests that are going to run on the same containers later. Complexity that is not worth for such a small test. I did check it locally though and error brought up as expected.

laurent-indermuehle (approved):
Thanks for the modification @ronger4! It's perfect now :)

Andersson007:
@ronger4 thanks for the contribution!
@laurent-indermuehle thanks for reviewing and the very valuable comments that made the new module much better!

PR #825, opened 2026-06-16
feat(mysql_clone): add mysql clone module

ronger4:
#### SUMMARY
- add a new `ansible.mysql.mysql_clone` module for MySQL Clone plugin operations that validates prerequisites, starts the clone, and waits for terminal completion
- add unit and integration coverage for clone support, including MariaDB rejection and predictive check mode behavior
- run the happy-path integration against a managed standalone mysql service because clone restarts the recipient instance and requires a supervised service to come back cleanly

#### ISSUE TYPE
- Feature Pull Request

#### COMPONENT NAME
- `plugins/modules/mysql_clone.py`
- `tests/unit/plugins/modules/test_mysql_clone.py`
- `tests/integration/targets/test_mysql_clone/defaults/main.yml`
- `tests/integration/targets/test_mysql_clone/meta/main.yml`
- `tests/integration/targets/test_mysql_clone/tasks/main.yml`
- `meta/runtime.yml`

#### ADDITIONAL INFORMATION
- verified with `ansible-test sanity` and `ansible-test units` for `mysql_clone`
- verified with focused MySQL integration runs for `test_mysql_clone`
- the integration target installs the Clone plugin only as test fixture setup; the module itself only validates plugin availability
- test_mysql_clone starts a standalone mysql-server service on the test host because the Clone plugin restarts the recipient MySQL instance as part of a successful clone. The default replica test containers run mysqld as a raw container process without a supervisor, which causes restart to fail (Restart server failed (mysqld is not managed by supervisor process)). Using a managed service gives the recipient a restartable MySQL instance, so the integration test can verify a real successful clone to Completed instead of accepting a timeout.
- MariaDB is not supported by this module because mysql_clone is a thin wrapper around MySQL’s native Clone plugin (CLONE INSTANCE FROM, clone plugin checks, and clone status tables in performance_schema). MariaDB does not expose an equivalent native feature in its official server docs. Supporting MariaDB would require a separate backup/provisioning implementation, likely around external tools such as mariadb-backup
- Used GPT5.4 with human in the loop and manual tests.

### DIFF
--- a/plugins/modules/mysql_clone.py
+++ b/plugins/modules/mysql_clone.py
@@ -41,6 +41,7 @@ options:
   donor_password:
     description:
     - Password used by the recipient to connect to the donor for clone.
+    - The password is sent to MySQL as part of the C(CLONE INSTANCE FROM ...) statement.
     type: str
     required: true
   require_ssl:
@@ -192,6 +193,19 @@ def get_redacted_query(query, params):
 
 def is_terminal_state(state):
     return state in TERMINAL_STATES
+
+
+def should_wait_after_execute_error(error_message):
+    if error_message is None:
+        return False
+
+    if 'Restart server failed' in error_message:
+        return True
+
+    if error_message.startswith('(3707,'):
+        return True
+
+    return False
 
 
 def _close_connection(cursor, connection):
@@ -361,6 +375,11 @@ def main():
             cursor.execute(query, params)
         except Exception as exc:
             execute_error = to_native(exc)
+            if not should_wait_after_execute_error(execute_error):
+                module.fail_json(
+                    msg='Clone failed to start. %s' % execute_error,
+                    query=redacted_query,
+                )
     finally:
         _close_connection(cursor, connection)
 
--- a/plugins/modules/mysql_clone.py
+++ b/plugins/modules/mysql_clone.py
@@ -79,6 +79,10 @@ attributes:
     details:
       - The module is not idempotent because each successful invocation starts a new clone operation.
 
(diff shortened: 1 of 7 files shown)


ronger4:
@Andersson007 @laurent-indermuehle can you review please?

Mariadb is not supported here due to the fact it requires external tools and its not out of the box.

laurent-indermuehle:
Hi @ronger4, I can't review this. I only manage MariaDB databases. I hope other maintainers can step-in?

plugins/modules/mysql_clone.py
```diff
@@ -0,0 +1,413 @@
...
+short_description: Clone a MySQL instance from a donor server
+
+description:
+- Starts a MySQL remote clone operation on the recipient server and waits until the clone reaches a terminal state.
+- Validates MySQL Clone plugin prerequisites on the recipient server but does not install or manage the plugin.
+- The module is limited to MySQL Clone plugin behavior and fails on MariaDB.
+
+author:
+- Ron Gershburg (@rgershbu)
+
```
Andersson007, line 24:
```suggestion

version_added: '5.1.0'

```

Andersson007 (changes requested):
@ronger4 thanks for the new module!
Please check out the review in 
[pr-825-review.txt](https://github.com/user-attachments/files/29280378/pr-825-review.txt)

Please let us know when it's ready for another round

Andersson007 (approved):

Andersson007:
@ronger4 thanks for the new module!

PR #829, opened 2026-06-18
ci: drop PyMySQL 0.9.3 and 1.0.2 from test matrix

Andersson007:
##### SUMMARY
  Drop PyMySQL 0.9.3 (unmaintained, unfixed CVE-2024-36039) and 1.0.2 (redundant with 1.1.1) from the
  CI matrix. Promote 0.10.1 to the main matrix.

  The remaining two versions cover both module code paths: < 1.0.0 (0.10.1) and >= 1.0.0 (1.1.1).

  Also removes dead test conditionals that checked for versions we no longer test against (< 0.10.0, >= 0.9.3, >= 0.9).

### DIFF
--- a/.agents/skills/run-tests/SKILL.md
+++ b/.agents/skills/run-tests/SKILL.md
@@ -63,14 +63,14 @@ Integration tests live under `tests/integration/targets/<module_name>/`. Each ta
 The project provides a Makefile (documented in `TESTING.md`) that handles spinning up a primary database and two replicas in Podman containers. This is the recommended way to run integration tests locally:
 
 ```bash
-# Run all targets against MySQL 8.0.38 with pymysql 1.0.2
-make ansible="stable-2.17" db_engine_name="mysql" db_engine_version="8.0.38" connector_name="pymysql" connector_version="1.0.2"
+# Run all targets against MySQL 8.4.9 with pymysql 1.1.1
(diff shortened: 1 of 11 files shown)


Andersson007:
ready for review

Andersson007:
resolved a merge conflict in doc files after merging another PR. ready for review

laurent-indermuehle:
@Andersson007 I see [new releases](https://github.com/PyMySQL/PyMySQL/releases) of PyMySQL added support for new Python:
-  Python 3.13 in v1.1.2

We didn't check this in our test matrix. I believe this wasn't an issue because we use the Python inside ubuntu-22.04 for all integrations tests.

Also 1.2.0 was released so maybe we should keep:
- 0.10.1
- 1.2.0

Andersson007:
> @Andersson007 I see [new releases](https://github.com/PyMySQL/PyMySQL/releases) of PyMySQL added support for new Python:
> 
> * Python 3.13 in v1.1.2
> 
> We didn't check this in our test matrix. I believe this wasn't an issue because we use the Python inside ubuntu-22.04 for all integrations tests.
> 
> Also 1.2.0 was released so maybe we should keep:
> 
> * 0.10.1
> * 1.2.0

Testing against 1.2.0 makes sense though there are a few breaking changes there, so it probably won't be just a bump. I would plan it as a separate work. I.e. I would keep this PR as-is for now and will create an issue for 1.2.0.
WDYT?

laurent-indermuehle:
Fine with me. Let's keep focused PR @Andersson007!

Andersson007:

